PR #15, opened 2015-06-02
CP-641 Listen for Request Cancellation in Client

evanweible-wf:
## Issue
- Request cancellation should cause the request to fail (throw an error async) in all scenarios. Currently, this works if the request is cancelled before being sent or after the response has been returned. If request cancellation occurs while the request is open, the request is cancelled but it does not result in an error being thrown.

## Changes
**Source:**
- Listen to `request.onAbort` in the client implementation to capture request cancellation while the request is open and throw a `WHttpException`.

**Tests:**
- One test updated, examples updated to utilize this new behavior.

## Areas of Regression
- Request cancellation.

## Testing
- Tests pass.
- In the File Transfer example, verify that canceling an upload and canceling a download still triggers the visual feedback (red background) that it did before. Previously, the visual feedback was triggered manually, but now it listens for an error to be thrown and then triggers the visual feedback. If this works as expected, it proves that this change works.

## Code Review
@maxwellpeterson-wf 

fyi: @trentgrover-wf

### DIFF
--- a/example/http/cross_origin_file_transfer/services/file_transfer.dart
+++ b/example/http/cross_origin_file_transfer/services/file_transfer.dart
@@ -67,8 +67,7 @@ class FileTransfer {
   /// Cancel the request (will do nothing if the request has already finished).
   void cancel(String reason) {
     _cancelled = true;
-    _request.abort();
-    _doneCompleter.completeError(reason != null ? new Exception(reason) : null);
+    _request.abort(reason != null ? new Exception(reason) : null);
   }
 
   void _progressListener(WProgress progress) {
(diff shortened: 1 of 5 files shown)


maxwellpeterson-wf:
+1 looks good

evanweible-wf:
@jayudey-wf could you look this over as well? If it looks good to you, go ahead with testing.

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - verified that in the cross origin file transfer example the red feedback was being displayed on cancelled downloads and uploads
- [ ] All unit tests pass

Merging into master.

PR #18, opened 2015-06-10
CP-664 Validate data type when sending, not before

evanweible-wf:
## Issue
- We should not immediately validate the data type; should validate it just before sending, instead.
- Without this change, it's not possible to set an intermediate version of a data payload (like w_service does) that may eventually get transformed into a valid data type

## Changes
**Source:**
- Move the data type validation step from the setter of the `data` property to the `_send` method. This allows the data to be set to anything, but only valid data types will actually be sent. Invalid data types will still throw an exception, just a bit later.

**Tests:**
- Tests updated.

## Areas of Regression
- Request data payload type validation

## Testing
- Tests pass.

## Code Review
@trentgrover-wf
@maxwellpeterson-wf

### DIFF
--- a/lib/src/http/w_http.dart
+++ b/lib/src/http/w_http.dart
@@ -231,12 +231,7 @@ class WRequest extends Object with FluriMixin {
   ///
   ///   - `Stream`
   ///   - `String`
-  void set data(Object data) {
-    common.validateDataType(data);
-    _data = data;
-  }
-  Object get data => _data;
-  Object _data;
+  Object data;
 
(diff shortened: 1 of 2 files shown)


maxwellpeterson-wf:
+1

evanweible-wf:
@jayudey-wf ready for merge.

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - tests pass
- [ ] All unit tests pass

Merging into master.

PR #41, opened 2015-07-08
CP-979 Add mocks for WRequest and WResponse

evanweible-wf:
## Issue
- We should provide mocks for the classes supplied in w_transport to make testing easier for consumers
- #32 

## Changes
**Source:**
- `WRequest` and `WResponse`
  - Break logic into more manageable pieces in order to make mocking easier
- `WRequestSource` and `WResponseSource`
  - Copy of original classes, minus some small pieces that are meant to be mocked out
  - Ideally, these should be generated
- `MockWRequest` and `MockWResponse`
  - Extends from the `Source` classes above
  - Mocks out a small selection of functionality
  - Adds an API for controlling the behavior

**Tests:**
- Add tests for the controllable behavior of the mocks

## Areas of Regression
- Sending of requests (existing tests should cover)

## Testing
- Tests pass
- The mocks should be easy to use in testing w_service (I will test this)

## TODO
- [ ] Mock for `WHttp`

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@jayudey-wf 

I mainly want to get feedback on this approach. I'm not convinced this is the best solution, at least not until I can research the possibility of generating the `Source` classes in more depth.

Issue #32, opened 2015-06-19
Provide mocks to make testing easier for consumers


evanweible-wf:
Things to consider:
- Avoid a dependency on mockito if possible since this will be included in the package source

PR #51, opened 2015-08-11
CP-819 Refactor internal source code organization

evanweible-wf:
## Issue
In working on mocks for our the classes exposed in this library, the need for a reorganization of the source code became apparent. We are already sharing a good amount of code between the client and server implementations, but there wasn't an easy way to extend that reuse to the mocks that we were attempting to create (#32 / #41). Additionally, the configuration for client or server usage - despite being easy for consumers - is messy internally. It currently consists of a bunch of variables that are null by default that then get populated with a platform-specific function.

## Changes
**Source:**
Now, the source code will be broken up into 5 different segments (seems like a lot, but it's much cleaner):

- **Public abstract classes.** Defines the API cleanly without any implementation code. Includes a factory constructor if the class allows construction.
  - `lib/src/http/w_http.dart`
  - `lib/src/http/w_http_exception.dart` __*__
  - `lib/src/http/w_progress.dart` __*__
  - `lib/src/http/w_request.dart`
  - `lib/src/http/w_response.dart`
  - * These two classes are small enough and don't depend on the platform that they only need one implementation, so that implementation lives in this file.
- **Common code.** Shared code that can remain the same on the client, server, or in a mock setting.
  - `lib/src/http/common/w_http.dart`
  - `lib/src/http/common/w_request.dart`
  - `lib/src/http/common/w_response.dart`
- **Platform-specific implementations.** Client/Server/Mock.
  - `lib/src/[client|server|mock]/w_http.dart`
  - `lib/src/[client|server|mock]/w_request.dart`
  - `lib/src/[client|server|mock]/w_response.dart`

With this organization in place, all that's necessary is to define a platform adapter that can construct the correct instances of `WHttp` and `WRequest` based on the configured platform.
> The public API for configuring w_transport remains the same to maintain backwards compatibility, but it now uses these new platform adapters internally.

**Tests:**
- Tests have been updated. A few tests were removed, but I verified that the content was also covered by integration tests.

## Areas of Regression
- Sort of everything..

## Testing
- CI build passes, examples work as expected.

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@dustinlessard-wf
fyi: @jayudey-wf


lib/src/mock_platform_adapter.dart
```diff
@@ -0,0 +1 @@
+
```
trentgrover-wf: intending to leave this in and complete the mock story?  or should this be deleted to handle the mock piece separately?

trentgrover-wf: it would definitely be nice to see how the mock adaptor fits into this, but I'm fine with handling this in 2 pieces if you want

trentgrover-wf:
+1

maxwellpeterson-wf:
+1

test/w_http_client_integration_test.dart
```diff
@@ -106,5 +106,27 @@ void main() {
       Map data = JSON.decode(await response.asText());
       expect(data['headers']['x-configured'], equals('true'));
     });
+
+    group('should set the withCredentials flag', () {
```
dustinlessard-wf: I'm not sure I understand this test.  
I expected to see an assert statement or expects statement.
Can you explain a little around what this is testing and how it works?

lib/src/mock_platform_adapter.dart
evanweible-wf: I didn't mean to include this, I think I'll tackle this separately

test/w_http_client_integration_test.dart
evanweible-wf: The `expect` statement is kind of hidden inside the `request.configure()` blocks below

dustinlessard-wf: Ok,I see now - thanks

dustinlessard-wf:
+1

evanweible-wf:
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - verified that examples were still working as expected
- [ ] All unit tests pass

Merging into master.

teresarevious-wf:
@rmconsole-wf

Issue #48, opened 2015-08-04
Add support for headers and body as optional params when sending requests via `WHttp` or `WRequest`


evanweible-wf:
Currently the API is a bit too restrictive and only allows setting the uri when dispatching a request. We should add support for headers and body via optional params for convenience.

evanweible-wf:
Fixed as of #67

PR #64, opened 2015-09-15
Adding support for sending headers as an optional param via WHttp and WRequest

jayudey-wf:
This is dependent on https://github.com/Workiva/w_transport/pull/61

## Issue
- #48 (appears that body / data is already an optional param)

## Changes
**Source:**
- modified parameters to be named
- added headers as an optional param

**Tests:**
- added tests to verify that headers can be set

## Areas of Regression
- WHttp and WRequest

## Testing
- ensure tests pass
- examples still work as expected


evanweible-wf:
Sorry, I forgot I was commenting on a commit :cry:

maxwellpeterson-wf:
I don't like the idea of changing `data` to `body`, it seems like every request library I've used in other languages have used `data`.

But as you said it seems like other Dart libs don't agree, so up to the rest of you on if you want to match that trend.

trentgrover-wf:
on @evanweible-wf 's discussion points:
- header replace vs merge -> I'd vote for merge
- on data vs body -> I lean towards keeping it as data, but am not too opinionated about it.  @dustinlessard-wf what's it look like in the Flex code base?

maxwellpeterson-wf:
@evanweible-wf perhaps both?

I'm not sure how you were planning on adding more native support for JSON conversion, but perhaps `body` could give you direct access to setting the body and `data` (or similar) could expect a `Map` and convert it to JSON for you automatically?

evanweible-wf:
@maxwellpeterson-wf I was planning on being more explicit and including `json` in the getter name. To me, having both `body` and `data` would be confusing, especially if they're different.

Maybe some more details about the response will help.. I'm hoping to include getters for the response body in 3 different formats:
- bytes
  - VM: response body is already stored as bytes
  - Browser: encoded version of the response body
- text
  - VM: decoded version of the response body
  - Browser: response body is already stored as text
- json
  - VM & browser: attempted JSON decode the text version of the response body

With that in mind, the naming might look something like...

```dart
// body
Uint8List get bodyBytes;
String get body;
Object get bodyJson;

// data
Uint8List get dataBytes;
String get data;
Object get dataJson;
```

or maybe..

```dart
// body
Uint8List get bytes;
String get body;
Object get json;

// data
Uint8List get bytes;
String get data;
Object get json;
```

The more I work through it in my head, the more I lean towards body.. but I'm still not convinced.
@trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf @jayudey-wf

trentgrover-wf:
I definitely wouldn't want both `body` and `data` with different meanings.  I'd rather have a syntax more like:
```
data.asString();
data.asBytes();
data.asJson();
```

evanweible-wf:
Okay, @jayudey-wf don't worry about `data` vs `body` in this PR, I'll address it in my request/response refactor branch.

trentgrover-wf:
FYI - I'm not even going to try to review this until #61 is merged and this diff is readable

jayudey-wf:
was it settled that we should merge headers rather than replace?

evanweible-wf:
@jayudey-wf Yeah, go ahead with merging headers :+1:

jayudey-wf:
closing PR due to all of the side conversation. will open separate PR after the #61 has been merged

rmconsole2-wf:
@jayudey-wf This pull request has merge conflicts, please resolve.

PR #66, opened 2015-09-23
CP-1024 Adding support for sending headers as an optional param via WHttp and WReques

jayudey-wf:
This is dependent on https://github.com/Workiva/w_transport/pull/61

## Issue
- #48 (appears that body / data is already an optional param)

## Changes
**Source:**
- modified parameters to be named
- added headers as an optional param

**Tests:**
- added tests to verify that headers can be set

## Areas of Regression
- WHttp and WRequest

## Testing
- ensure tests pass
- examples still work as expected

## Code Review
@dustinlessard-wf 
@evanweible-wf 
@maxwellpeterson-wf 
@trentgrover-wf


lib/src/http/w_request.dart
```diff
@@ -56,12 +54,18 @@ abstract class WRequest implements FluriMixin {
   /// [WProgress] stream for this HTTP request's download.
   Stream<WProgress> get downloadProgress;
 
+  /// Map of HTTP request headers
+  CaseInsensitiveMap get headers;
```
evanweible-wf: This public type should still be `Map<String, String>`. It will still work since `CaseInsenstiveMap<String> implements Map<String, String>`.

lib/src/mocks/http.dart
```diff
@@ -156,10 +179,12 @@ class MockHttp {
...
 class _RequestExpectation {
   Object failWith;
+  final CaseInsensitiveMap headers;
   final String method;
   WResponse respondWith;
   final Uri uri;
 
-  _RequestExpectation(String this.method, Uri this.uri,
+  _RequestExpectation(
+      CaseInsensitiveMap this.headers, String this.method, Uri this.uri,
```
evanweible-wf: Could we move `headers` to the third positional parameter? I know this is alphabetized, but with positional args, order matters, and method -> uri -> headers seems like the most logical ordering.

Also the type should be `Map<String, String>` here - it will be converted to a `CaseInsenstiveMap` internally

lib/src/mocks/http.dart
```diff
@@ -94,15 +111,21 @@ class MockHttp {
...
       throw new ArgumentError('Use failWith OR respondWith, but not both.');
     }
     if (failWith == null && respondWith == null) {
       respondWith = new MockWResponse.ok();
     }
-    _expectations.add(new _RequestExpectation(method, uri,
-        failWith: failWith, respondWith: respondWith));
+    _expectations.add(new _RequestExpectation(
+        headers == null
+            ? new CaseInsensitiveMap()
```
evanweible-wf: Converting a null value for `headers` to an empty map ruins the check at line 46 above. The idea is that if `headers` is null, then the tester doesn't care about them and will match requests based only on URI and method. By converting this to an empty map, the tester won't match requests that have headers..

```dart
MockTransports.http.expect('GET', Uri.parse('/path'));
WHttp.get(Uri.parse('/path'), headers: {'x-custom': 'value'});
```
I would expect this request to resolve because my `.expect()` call didn't specify headers.

```dart
MockTransports.http.expect('GET', Uri.parse('/path'), headers: {});
WHttp.get(Uri.parse('/path'), headers: {'x-custom': 'value'});
```
I would expect this request not to resolve since I explicitly stated an expectation of no headers.

evanweible-wf:
Couple small comments, otherwise looks good!

trentgrover-wf:
+1

evanweible-wf:
+1

jayudey-wf:
could someone +10 this one too?

rmconsole2-wf:
@jayudey-wf This pull request has merge conflicts, please resolve.

PR #69, opened 2015-11-02
CP-1089 Allow headers, withCredentials to be set on Client

evanweible-wf:
## Issue
- To reduce repetition (and to make `Client` more useful), you should be able to set headers and the `withCredentials` flag on a `Client` and have those applied to all requests created by that client.

## Changes
**Source:**
- Add `headers` and `withCredentials` getters/setters to `Client`
- Apply values to all requests created by `Client`

**Tests:**
- Unit tests added to verify this `Client` behavior

## Areas of Regression
- n/a, new functionality

## Testing
- CI passes

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@dustinlessard-wf
@jayudey-wf

### DIFF
--- a/lib/src/mocks/http.dart
+++ b/lib/src/mocks/http.dart
@@ -16,6 +16,7 @@ library w_transport.src.mocks.http;
 
 import 'dart:async';
 
+import 'package:http_parser/http_parser.dart' show CaseInsensitiveMap;
 import 'package:w_transport/w_transport.dart';
 
 import 'package:w_transport/src/http/finalized_request.dart';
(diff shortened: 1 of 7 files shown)


maxwellpeterson-wf:
+1

lib/src/mocks/http.dart
```diff
@@ -36,8 +37,27 @@ void cancelMockRequest(MockBaseRequest request) {
...
+    bool uriMatches = e.uri == request.uri;
+    bool headersMatch;
+    if (e.headers == null) {
+      // Ignore headers check if expectation didn't specify.
+      headersMatch = true;
+    } else if (e.headers.isEmpty) {
+      headersMatch = request.headers.isEmpty;
+    } else {
+      headersMatch = true;
+      e.headers.forEach((header, value) {
```
trentgrover-wf: technically, if `request` includes headers that aren't included in `e`, this will return as a successful match, even though they are not actually a match.  Does that matter for the usage here?

evanweible-wf: Yeah that's what I wanted because some default headers get added (`content-type`, `content-length`). If the tester wants to verify those as well, they can be including them in the expectation, but I thought it would be annoying to have to include headers that aren't explicitly set

trentgrover-wf: sounds fine

trentgrover-wf:
+1

trentgrover-wf:
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI
- [ ] All unit tests pass

Merging into master.

chadknight-wf:
+1 security

Issue #60, opened 2015-09-04
WebSocket: Exception thrown as result of call to addError() should be catchable


evanweible-wf:
Things to consider:
- The exception that gets thrown as a result of calling `addError()` should be catchable by listening to the `done` future.
- The server WebSocket class has an `addError()` method, so we can forward it, but the client WebSocket class does not, so behavior is inconsistent. It might be better to handle the error ourselves and close the socket instead of trying to pass the error into the underlying WebSocket class.
- There are two `TODO` tests that should pass if this behavior is working as expected

evanweible-wf:
Fixed in #71

PR #71, opened 2015-11-04
CP-1092 Cleanup WSocket, handle errors correctly, naming changes

evanweible-wf:
## Issue
- Need to conform to new naming conventions (Client --> Browser, Server --> VM)
- `addError()` was not behaving correctly - could not catch the exception
- Not confident in how we are monitoring the sink/socket status and reacting to them closing in correct order

## Changes
**No breaking changes!** All of this is an refactor to the internal code - no public API changes required.

**Source:**
- `Client` --> `Browser`
- `Server` --> `VM`
- `WSocket.addError()` now handles the error itself instead of passing it on to the underlying socket. Adding an error to the socket would just cause it to close, so we explicitly close it now instead. This allows us to control when the error is thrown, allowing consumers to catch it via the `WSocket.done` Future.
- Quite a bit of the `WSocket` logic was moved up into the `CommonWSocket` class, reducing the platform-dependent code and DRYing up the code.
- All implementations of `WSocket` now handle the outgoing sink and incoming socket connection in a much more clean manner. Centralized some logic into a `shutDown()` method that handles closing the sink and socket. We explicitly wait for both the sink and the socket to finish closing before finalizing some info and completing the `done` Future.
- `MockWSocket` class updated to better align with the real implementations.

**Tests:**
- `addError()` integration test now passes!

## Areas of Regression
- WebSocket (`WSocket` class)

## Testing
- CI passes
- Socket example passes

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@dustinlessard-wf
@jayudey-wf


evanweible-wf:
This addresses #60

lib/src/web_socket/common/w_socket.dart
```diff
@@ -27,20 +32,71 @@ abstract class CommonWSocket extends Stream implements WSocket {
...
+
+  /// Outgoing communication. Sink will be exposed, allowing users to
+  /// add items to the outgoing stream.
+  StreamController outgoing;
+
+  /// The close event with close code and reason from the socket connection.
+  /// Will only be populated once the socket has actually closed.
+  WSocketCloseEvent _closeEvent;
+
+  /// Completer that will complete when [_allClosed] has completed and the close
```
trentgrover-wf: this comment uses `_allClosed` but the var above is `allClosed` (not private)

lib/src/web_socket/common/w_socket.dart
```diff
@@ -27,20 +32,71 @@ abstract class CommonWSocket extends Stream implements WSocket {
...
+  Completer<Null> _done = new Completer();
 
   CommonWSocket() {
-    done.whenComplete(() {
-      _closed = true;
+    // Wait for both the sink and the socket to finish closing before exposing
+    // the close code, close reason, and completing the [_done] completer.
+    // If the socket closed due to an error, the [_done] completer will be
+    // completed with that error.
+    allClosed.future.then((_) {
```
trentgrover-wf: is there a reason for `allClosed` to be public?  I would suspect that there could be async issues if a consumer waited for `allClosed` instead of `done` because there's a tiny risk that these `closeCode` / `closeReason` vals might not yet be valid

lib/src/web_socket/mock/w_socket.dart
```diff
@@ -33,42 +34,50 @@ abstract class MockWSocket implements WSocket {
...
 class _MockWSocket extends CommonWSocket implements MockWSocket, WSocket {
   List<Function> _callbacks = [];
 
-  Completer _done = new Completer();
+  int _code;
 
   List<Function> _errorCallbacks = [];
 
-  StreamController _sinkController = new StreamController();
+  StreamController _mocket = new StreamController();
```
trentgrover-wf: my inner wocket :heart: mocket

lib/src/web_socket/common/w_socket.dart
evanweible-wf: Changed variable back to _allClosed.

lib/src/web_socket/common/w_socket.dart
evanweible-wf: I had then public because I thought the extending classes need them, but I ended up moving more code than I originally thought into the common class, so only `incoming` and `outgoing` need to be public for that reason. The rest I've switched back to private

evanweible-wf: Also these being public doesn't actually expose them to consumers. Only the fields defined in the `WSocket` abstract class are available

trentgrover-wf:
+1

lib/src/web_socket/mock/w_socket.dart
```diff
@@ -30,45 +31,49 @@ abstract class MockWSocket implements WSocket {
...
 
-// TODO expose a public interface for this class to limit the API to what would only be useful for testing
 class _MockWSocket extends CommonWSocket implements MockWSocket, WSocket {
   List<Function> _callbacks = [];
 
-  Completer _done = new Completer();
+  int _code;
 
-  List<Function> _errorCallbacks = [];
+  StreamController _mocket = new StreamController();
```
maxwellpeterson-wf: :smile:

maxwellpeterson-wf:
+1

maxwellpeterson-wf:
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - CI passes, socket example still works as expected
- [ ] All unit tests pass

Merging into master.

PR #67, opened 2015-09-29
CP-1058 HTTP API redesign, w_transport

evanweible-wf:
> **There are still pieces of this that are in progress:**
- [x] Inline documentation
- [x] Readme updates (may be deferred until #49)
- [x] Tests (existing ones are passing, but new ones need to be added for new functionality)

## Issue
The current HTTP API is limited and does not support the various types of requests well.
- Support for various request body types is inconsistent across platforms.
- No explicit support for the following request types:
  - JSON (application/json)
  - Form (application/www-form-urlencoded)
  - Multipart (multipart/form-data)
- The response body is not available synchronously (#40)
- The response body must be manually decoded to JSON (#44)
- Support for content-type, charset, and encoding is essentially manual.

## Changes
**Source:**
_TODO_.

**Tests:**
- Existing tests have been updated and are passing.
- New tests need to be added.

## Areas of Regression
- Everything.

## Testing
TBD

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@dustinlessard-wf
@jayudey-wf 

fyi: @brentarndorfer-wf @tylertreat @stevenosborne-wf @markerickson-wf

PR #72, opened 2015-11-04
CP-1093 Increase HTTP coverage. Consistent test naming.

evanweible-wf:
## Issue
- Still a few gaps in test coverage around HTTP code
- Integration tests have a pattern for test names but the rest of the tests don't follow this

## Changes
**Source:**
- Fixed an issue where types were being checked in the wrong order (`Blob` vs `File`)
- Browser requests (XHR) now set `responseType = 'blob'` for responses that should be streamed and a `FileReader` is used to obtain the response in bytes. This prevents the browser from having to decode the response which previously was immediately being encoded back into bytes for this use case.

**Tests:**
- Additional tests around HTTP code
- Made test naming consistent for all tests, now includes:
  - test type (unit, integration)
  - topic (HTTP, WebSocket, Mocks, PlatformAdapter)
  - platform (Browser, VM, Mock - integration tests only).

## Areas of Regression
- File Upload (covered by file transfer example)
- Streamed responses (covered by tests)

## Testing
- CI passes
- File transfer example works as expected

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@dustinlessard-wf
@jayudey-wf


evanweible-wf:
The diffs look worse than they are.. I recommend viewing them without whitespace changes: https://github.com/Workiva/w_transport/pull/72/files?w=0

dustinlessard-wf:
Only 67 files changed... damn, was hoping for a high-score. :-)

evanweible-wf:
only about 100 files short of #67  :smile:

maxwellpeterson-wf:
+1 nice!

test/naming.dart
```diff
@@ -19,19 +19,27 @@ const String platformBrowser = 'browser';
...
+const String topicMocks = 'Mocks';
+const String topicPlatformAdapter = 'Platform Adapter';
+const String topicWebSocket = 'WS';
+
+class Naming {
+  String platform;
+  String testType;
+  String topic;
+
+  @override
```
dustinlessard-wf: I was curious if one must include `@override` with the toString function.  It turns out that it's not necessary : https://dartpad.dartlang.org/053c01d956e344df2f67
No biggie including it though.

dustinlessard-wf:
+1, I found myself quite a few times thinking : "Oh cool, I like that - I think I'm going to copy that approach"

test/naming.dart
evanweible-wf: It's actually never necessary, the benefit you get from it is if you include it on something that doesn't actually override something from a parent class. So if you have a typo or if you refactor something in a parent class and forget to change the child class, the compiler can warn you.

dustinlessard-wf: Thanks - that makes sense.  I had a feeling you had a reason for including it!

lib/src/http/browser/request_mixin.dart
```diff
@@ -81,9 +80,10 @@ abstract class BrowserRequestMixin implements BaseRequest, CommonRequest {
         c.complete(_createResponse(streamResponse: streamResponse));
       }
     });
-    void onError(error) {
+    onError(error) async {
```
trentgrover-wf: might as well toss a return type on this

trentgrover-wf:
+1

lib/src/http/browser/request_mixin.dart
evanweible-wf: added

trentgrover-wf:
+1

maxwellpeterson-wf:
+1

evanweible-wf:
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - tests pass, file download example works as expected
- [ ] All unit tests pass

Merging into master.

PR #92, opened 2015-12-09
CP-1177 Support request, response interception

evanweible-wf:
## Issue
- To support more advanced functionality in an http `Client`, we need a way to intercept requests right before dispatch and responses right before delivery.

## Changes
**Source:**
- Add a `requestInterceptor` field to `BaseRequest` that, if set, will be applied right before dispatching the request.
  - Signature:

    ```
    Future<Null> RequestInterceptor(BaseRequest request)
    ```

- Add a `responseInterceptor` field to `BaseRequest` that, if set, will be applied right after the response is received but before it's delivered to the caller.
  - Signature:
    
    ```
    Future<BaseResponse> ResponseInterceptor(
        FinalizedRequest request, BaseResponse response,
        [RequestException error])
    ```

- Add a `Pathway<T>` class that defines a path on which a payload of type `T` travels and on which interceptors can be registered.

- Add an `HttpInterceptor` class as a way to encapsulate both request and response interception in a single class (with default implementations that simply return the given payload sans-modification)

- Add `RequestPayload` and `ResponsePayload` classes to encapsulate the information needed by request and response interceptors.

- Add an `addInterceptor()` method to `Client` to allow the registration of multiple interceptors that will be applied serially to every request created from the `Client` instance as well as their responses (thus accomplishing the original goal)

**Tests:**
- Unit tests added to cover all of the added functionality.

## Areas of Regression
- Requests when `requestInterceptor` and/or `responseInterceptor` are not set (should be covered by the request tests that already exist).

## Testing
- CI passes
- Examples function as expected

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@dustinlessard-wf
@jayudey-wf

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,14 +18,15 @@
   - [Dart VM](#dart-vm)
   - [Tests](#tests)
 - [**HTTP**](#http)
-  - [Simple Requests](#simple-requests)
-  - [Advanced Requests & Common Request API](#advanced-requests)
+  - [Static Request Methods](#static-request-methods)
+  - [Request Classes](#request-classes)
+  - [Common Request API](#common-request-api)
     - [Creating a Request](#creating-a-request)
-    - [Bytes](#bytes)
-    - [Content-Length, Content-Type and Encoding](#content-length-content-type-and-encoding)
     - [Canceling a Request](#canceling-a-request)
-    - [Timeout Threshold](#timeout-threshold)
     - [Credentials (browser only)](#credentials-browser-only)
+    - [Content-Length, Content-Type and Encoding](#content-length-content-type-and-encoding)
+    - [Timeout Threshold](#timeout-threshold)
+    - [Request & Response Interception](#request--response-interception)
   - [Request Types](#request-types)
     - [JsonRequest](#jsonrequest)
     - [FormRequest](#formrequest)
@@ -33,6 +34,7 @@
     - [Request (plain-text)](#request-plain-text)
     - [StreamedRequest](#streamedrequest)
   - [HTTP Client](#http-client)
+    - [Intercepting Requests & Responses from a Client](#intercepting-requests--responses-from-a-client)
   - [Responses](#responses)
   - [Streamed Responses](#streamed-responses)
 - [**WebSocket**](#websocket)
@@ -112,7 +114,7 @@ void main() {
 
 ## HTTP
 
-### Simple Requests
+### Static Request Methods
 
 For one-off or simple requests, use the static methods on the `Http` class:
 ```dart
@@ -129,7 +131,6 @@ These standard HTTP methods are supported:
 * PATCH : `Http.patch()`
 * POST : `Http.post()`
 * PUT : `Http.put()`
-* TRACE : `Http.trace()` (forbidden in the browser)
 
 If you need to send a request with a non-standard HTTP method, use `send()`:
 ```dart
@@ -153,11 +154,11 @@ await Http.post(Uri.parse('/notes/'), body: 'testing..');
 > Plain-text request bodies default to UTF8 encoding.
 
 
-### Advanced Requests
+### Request Classes
 The above works well for simple requests, but what if you need to send JSON? Or
 use a different encoding? Send a multi-part request? Or maybe you just want more
 explicit control over the request. In these cases, you'll want to use one of the
-exposed `Request` classes:
+exposed request classes:
 
 * `Request`
 * `FormRequest`
@@ -165,11 +166,14 @@ exposed `Request` classes:
 * `MultipartRequest`
 * `StreamedRequest`
 
-#### Common Request API
+> Every request class extends from `BaseRequest`.
+
+
+### Common Request API
 Though each request type has some idiosyncrasies, they share a common underlying
 API. We'll cover this common API using `Request` as the example.
 
-##### Creating a Request
+#### Creating a Request
 ```dart
 Request request = new Request();
 ```
@@ -188,7 +192,7 @@ they can be set on the `Request` directly before being sent:
 
 ```dart
 Request request = new Request()
-  ..uri = Uri.parse('/notes/'),
+  ..uri = Uri.parse('/notes/')
   ..headers = {'x-auth-token': 'a390bn'}
   ..body = 'A note.';
 await request.post();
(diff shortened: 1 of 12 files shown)


lib/src/http/common/request.dart
```diff
@@ -62,6 +58,15 @@ abstract class CommonRequest extends Object
   /// HTTP method ('GET', 'POST', etc).
   String method;
 
+  /// Request interceptor. Called right before request is sent.
+  /// TODO: prevent modification after request sending has started.
+  RequestInterceptor requestInterceptor;
+
+  /// Response interceptor. Called after response is received and before it is
+  /// delivered to the request sender.
+  /// TODO: prevent modification after request sending has started.
```
evanweible-wf: Just saw these TODOs that I forgot about - I'm fixing it now

lib/src/http/common/request.dart
```diff
@@ -415,16 +446,23 @@ abstract class CommonRequest extends Object
...
       if (!_done.isCompleted) {
         _done.complete();
       }
       cleanUp();
-      checkForCancellation(response: response);
       if (error is! RequestException) {
         error = new RequestException(method, this.uri, this, response, error);
       }
+      if (responseInterceptor != null) {
+        response = await responseInterceptor(finalizedRequest, response, error);
```
maxwellpeterson-wf: It seems as if this `catch` block could be entered if there was an error encountered during `await responseInterceptor` on line 451, and then it would do it again on line 463. Is that a correct understanding? Should we prevent that situation?

lib/src/http/http_interceptor.dart
```diff
@@ -0,0 +1,73 @@
...
+/// 2. a set of interceptors can be attached.
+///
+/// As a payload travels a pathway, each interceptor (in the order they were
+/// attached) has the opportunity to process, evaluate, modify, augment, or
+/// replace the payload. This interception can be synchronous or asynchronous -
+/// the pathway ensures that each interceptor is applied serially.
+///
+/// Simply put, a pathway accepts a payload and asynchronously returns a
+/// payload.
+class Pathway<T> {
```
maxwellpeterson-wf: Wouldn't this be provided by the interceptor-dart library?

evanweible-wf: For now, it's small enough that I'm including it in here and eventually we can break it out.

lib/src/http/common/request.dart
evanweible-wf: Good catch (pun intended), there should probably be a try/catch around line 451 that would record that the error happened and then rethrow it.. then we can avoid calling it again and still wrap the error in a `RequestException`

evanweible-wf:
@maxwellpeterson-wf added logic to handle the response interceptor throwing. Also added tests to assert the expected behavior when either the request or the response interceptor throws.

maxwellpeterson-wf:
+1 nice

evanweible-wf:
@trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf updated the readme and changelog.

README.md
```diff
@@ -476,6 +509,79 @@ Client client = new Client();
...
+  // `baz=bar` query parameter.
+  client.newRequest().get(uri: Uri.parse('...'));
+```
+
+Obviously these examples are contrived, but this pattern enables some powerful
+functionality. Consider the following interceptors as possibilities:
+
+- **Analytics:** records types of requests, request duration, request
+  failures, etc.
+- **CSRF:** sets the `x-xsrf-token` header on outgoing mutation requests and
```
maxwellpeterson-wf: `x-xsrf-token` is somewhat Workiva specific. Maybe we want to say "set a header for CSRF verification" instead? I don't really mind either way, more of a note.

evanweible-wf: yeah probably no need to specify an actual header here

maxwellpeterson-wf:
+1

README.md
evanweible-wf: fixed

maxwellpeterson-wf:
+1

README.md
```diff
@@ -304,20 +309,44 @@ request.get();
...
 ```dart
-Request request = new Request()
-  ..uri = Uri.parse('https://otherhost.com/notes/')
-  ..withCredentials = true;
-await request.get();
+Request request = new Request();
+
+// Register a hook to intercept the request.
+request.requestInterceptor = (Request request) async {
+  // Modify the request as necessary.
```
trentgrover-wf: are you expecting that all request modifications will be made in place?  or should we expect that a `requestInterceptor` returns a request object instead?

can you think of any use cases where a consumer would want to return a brand new request object instead of modifying the existing?

README.md
```diff
@@ -304,20 +309,44 @@ request.get();
...
+Again, this interception can be async if necessary.
+
+```dart
+Request request = new Request();
+
+// Register a hook to intercept the response.
+request.responseInterceptor =
+    (FinalizedRequest request, BaseResponse response,
+    [RequestException exception]) async {
+  // Return a `BaseResponse` instance, modified as necessary.
```
trentgrover-wf: i'd vote for both interceptors to either modify the incoming request / response in place or explicitly return an object instance instead of request and response pattern being different

README.md
```diff
@@ -476,6 +509,79 @@ Client client = new Client();
...
+}
+
+main() {
+  Client client = new Client()
+    ..addInterceptor(new HeaderInterceptor())
+    ..addInterceptor(new QueryParamInterceptor());
+
+  // The client will create a request interceptor that chains together the logic
+  // from both of the interceptors registered above. This will be set on the
+  // request, meaning that this request will have an `x-baz: bar` header and a
```
trentgrover-wf: `x-baz` -> `x-foo`?

README.md
```diff
@@ -476,6 +509,79 @@ Client client = new Client();
...
+  // `baz=bar` query parameter.
+  client.newRequest().get(uri: Uri.parse('...'));
+```
+
+Obviously these examples are contrived, but this pattern enables some powerful
+functionality. Consider the following interceptors as possibilities:
+
+- **Analytics:** records types of requests, request duration, request
```
trentgrover-wf: you said it, now we have to do it... :)

lib/src/http/base_request.dart
```diff
@@ -20,8 +20,16 @@ import 'dart:convert';
 import 'package:fluri/fluri.dart';
 import 'package:http_parser/http_parser.dart';
 
+import 'package:w_transport/src/http/finalized_request.dart';
 import 'package:w_transport/src/http/request_dispatchers.dart';
+import 'package:w_transport/src/http/request_exception.dart';
 import 'package:w_transport/src/http/request_progress.dart';
+import 'package:w_transport/src/http/response.dart';
+
+typedef Future<Null> RequestInterceptor(BaseRequest request);
```
trentgrover-wf: opinions on making this `typedef Future<BaseRequest> RequestInterceptor(BaseRequest request);`?

trentgrover-wf:
@evanweible-wf looks awesome, just a couple fine details to debate

README.md
evanweible-wf: I was hoping to prevent interceptors from being able to return a different `BaseRequest` instance because that would create a disconnect between the original request (on which the caller would be listening) and the request that actually gets sent.

README.md
evanweible-wf: I agree that this is confusing, but I'm not sure what to do about it. I don't want to allow the request instance to be replaced, which leaves in-place modification; and `BaseResponse` instances can't be modified in place (fields are final), which means it has to be replaced..

lib/src/http/base_request.dart
evanweible-wf: I suppose this could work if I asserted that returned instance is the same instance as was passed in (I think `identical()` would work for that?). Would need to make sure the messaging/docs around that are clear.

README.md
evanweible-wf: Sounds good! But in all seriousness, I think it would be cool/make sense to create a repo with useful interceptors like this. Or maybe those just live in a separate entry point in this repo.

rmconsole2-wf:

README.md
evanweible-wf: Good catch, fixed.

evanweible-wf:
@trentgrover-wf I added a clarifying note to the readme after the request/response interception example as well as a note to the `BaseRequest.requestInterceptor` doc comment.

trentgrover-wf:
+1

maxwellpeterson-wf:
+1 - still has merge conflicts though

evanweible-wf:
@maxwellpeterson-wf yup, was waiting for +1s to rebase & squash

evanweible-wf:
@trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf rebased against latest master and squashed commits.

trentgrover-wf:
+1

maxwellpeterson-wf:
+1

README.md
dustinlessard-wf: Do we have an understanding for this point?

evanweible-wf: Sorry - we discussed offline. We're going to leave it as is and I added some clarifying documentation

dustinlessard-wf:
+1 despite the merge conflicts.

evanweible-wf:
@dustinlessard-wf merge conflicts were resolved

evanweible-wf:
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - verified that examples were working as expected
- [ ] All unit tests pass

Merging into master.

Issue #95, opened 2015-12-09
Support request retrying


evanweible-wf:
It would be useful if all requests and `Client` had an API to enable request retrying.

## Criteria
- There should be criteria (configurable) that determines whether or not a failed request should be retried based on the status code, request method, etc.
  - By default, mutation requests should not be retried to avoid issues with idempotence.
  - 500 and 502 should be retried because they could be transient.
- The number of retries to attempt should have a sensible default and be configurable.
- There should be an API for retry-related info like `bool get wasRetried`, `int get numRetries`, `Iterable get retryFailures`.
  - These should live on the request instance so that they are always available. This means it will also need to be on the `FinalizedRequest` class so that response interceptors have access.
  - Interceptors can use this information to prioritize requests or to record analytics.
- If a request exceeds the maximum number of retries, an explanation should be included in the string version of the `RequestException`

## Questions
- Should `timeoutDuration` apply to each request attempt or to the process as a whole?
  - Answer: each attempt

@trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf

maxwellpeterson-wf:
I vote each attempt.

evanweible-wf:
+1

PR #99, opened 2015-12-14
CP-1186 Automatic retrying for failed requests

evanweible-wf:
_Resolves #95_

## Issue
- #95 We should support configurable and transparent request retrying for failed requests

## Changes
**Source:**

### `AutoRetry`
- Add an `AutoRetry` class that encapsulates all configuration and information for a single request
  - Auto retry is opt-in (`enabled` flag must be set to `true`)
  - Max number of retries is configurable
  - Determining whether or not a request is retried is based on 3 factors (each of which has defaults and is configurable) that must each pass for the request to be eligible:
    - HTTP method
    - Response status code
    - Custom test (function that receives the finalized request and the response and returns true/false)
  - Contextual information is available (may be useful for interceptors to determine where in the lifecycle a request is, or for analytics, or for application code after a request finishes, etc)
    - # of attempts
    - list of failures
  - Auto retry is also not supported for certain requests
    - `StreamedRequest`s - can't read the stream body more than once
    - `MultipartRequest`s that contain files in the body - files' byte streams can't be read more than once

### `BaseRequest`
- Add a `clone()` method - necessary for requests
- `BaseRequest` provides a base implementation that constructs the correct request instance (taking into account whether or not it was created from a `Client`)
- Each implementing class of `BaseRequest` overrides this method, calls super to get the base clone, and then adds any request-specific details

### `CommonRequest`
- Support request retrying by testing a failed request for retry eligibility and attempting to retry a clone of the request. If the retry succeeds, that response is returned instead. If the retry eventually fails, that exception is returned.
- Switched to using a `Completer` for some of the async logic so that parts of the stack trace are not lost

### TODO
- [x] Add the `clone()` method to the browser implementation of `MultipartRequest` (currently missing)
- [x] Add a subset of the fields from the `AutoRetry` class to the `Client` class so that retry configuration can be easily applied to all requests
- [x] ~~Add some sort of logging/information when a request that would otherwise be eligible for retry is not retried because it is unsupported (see last bullet of `AutoRetry` above)~~ Deferring until https://github.com/Workiva/w_transport/issues/100
- [x] Update changelog

**Tests:**
- Unit tests added for several different variations of failed/successful requests with different auto-retry configurations.
- TODO:
  - [x] Test that request cancellation during a retry attempt functions correctly
  - [x] Add integration tests for request retrying
  - [x] Test `clone()` (base implementation and request-specific implementations)
  - [x] Test for attempting to retry streamed requests
  - [x] Test for attempting to retry multipart requests with files
  - [x] Test for requests created from a client and then cloned
  - [x] 100% coverage of added lines

## Areas of Regression
- HTTP request dispatch (should be adequately covered by tests)

## Testing
- CI passes
- Examples function as expected
- (Side note: I've created an issue to add an example that demonstrates request retrying https://github.com/Workiva/w_transport/issues/98)

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@dustinlessard-wf
@jayudey-wf

### DIFF
--- a/lib/src/http/auto_retry.dart
+++ b/lib/src/http/auto_retry.dart
@@ -1,0 +1,147 @@
+// Copyright 2015 Workiva Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+library w_transport.src.http.auto_retry;
+
+import 'dart:async';
+
+import 'package:w_transport/src/http/base_request.dart';
+import 'package:w_transport/src/http/finalized_request.dart';
+import 'package:w_transport/src/http/request_exception.dart';
+import 'package:w_transport/src/http/requests.dart';
+import 'package:w_transport/src/http/response.dart';
+
+typedef Future<bool> RetryTest(
+    FinalizedRequest request, BaseResponse response, bool willRetry);
+
+/// Deciding whether or not to retry a failed request is determined by the
+/// settings defined in fields in this class.
+///
+/// To turn automatic retrying on, set [enabled] to `true`.
+///
+/// The decision to retry a request or not can usually be made based solely on
+/// the request method and the response status code. For this reason, the
+/// [forHttpMethods] and [forStatusCodes] fields define a list of methods and
+/// status codes, respectively, for which retrying is acceptable.
+///
+/// There is also a [test] field for a custom test function that can make the
+/// decision based on more information.
+///
+/// If [test] is defined (not null), then it is called with:
+///
+/// 1. A [FinalizedRequest] instance,
+/// 2. A [BaseResponse] instance, and
+/// 3. A `willRetry` boolean representing whether or not the HTTP method and
+///    status code checks passed.
+///
+/// The [test] then makes the final decision (retry or no retry) by
+/// asynchronously returning a boolean.
+///
+/// If, however, [test] is null, then the decision is made based solely on
+/// whether or not the HTTP method and status code checks passed.
+class AutoRetryConfig {
+  /// Whether or not automatic retrying is enabled.
+  bool enabled = false;
+
+  /// The set of HTTP methods that are considered eligible for retrying.
+  ///
+  /// By default, non-mutable HTTP methods included here (GET, HEAD, OPTIONS).
+  /// This is to avoid issues with requests that may not be idempotent.
+  List<String> forHttpMethods = ['GET', 'HEAD', 'OPTIONS'];
+
+  /// The set of status codes that are considered eligible for retrying.
+  ///
+  /// By default, 500, 502, 503, and 504 are included here because they
+  /// represent server errors that may be transient.
+  List<int> forStatusCodes = [500, 502, 503, 504];
+
+  /// Maximum number of retries to attempt. This excludes the original request.
+  /// For example, a request with `maxRetries = 2` will produce up to 3 requests
+  /// total - the first request and 2 retries.
+  int maxRetries = 2;
+
+  /// A custom [test] function that decides whether or not a request should be
+  /// retried. It will be called with:
+  ///
+  /// 1. A [FinalizedRequest] instance,
+  /// 2. A [BaseResponse] instance, and
+  /// 3. A `willRetry` boolean representing whether or not the HTTP method and
+  ///    status code checks passed.
+  ///
+  /// The [test] then makes the final decision (retry or no retry) by
+  /// asynchronously returning a boolean.
+  ///
+  ///     // Example of auto retry with a custom check for a CSRF failure that
+  ///     // can only be identified by a message in the response body.
+  ///
+  ///     var request = new Request();
+  ///     request.autoRetry
+  ///       ..enabled = true
+  ///       ..test = (FinalizedRequest request, BaseResponse response,
+  ///           bool willRetry) async {
+  ///         // Check for a special case (CSRF failure) by reading the body.
+  ///         // If it's determined that it is a CSRF failure, return `true`
+  ///         // to indicate the request should be retried.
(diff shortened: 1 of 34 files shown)


rmconsole2-wf:

lib/src/http/auto_retry.dart
```diff
@@ -0,0 +1,105 @@
+library w_transport.src.http.auto_retry;
+
+import 'package:w_transport/src/http/base_request.dart';
+import 'package:w_transport/src/http/finalized_request.dart';
+import 'package:w_transport/src/http/request_exception.dart';
+import 'package:w_transport/src/http/requests.dart';
+import 'package:w_transport/src/http/response.dart';
+
+typedef /* Future<bool> | bool */ RetryTest(
+    FinalizedRequest request, BaseResponse response);
```
trentgrover-wf: no need for this distinction here.  just assume the return type is `Future<bool>`.  if you do that, the async keyword is all that's necessary for a simple function to match the signature:
```
Future<bool> tryIt() async {
  return false;
}
```

lib/src/http/auto_retry.dart
```diff
@@ -0,0 +1,105 @@
...
+
+  /// Maximum number of retries to attempt. This excludes the original request.
+  /// For example, a request with `maxRetries = 2` will produce up to 3 requests
+  /// total - the first request and 2 retries.
+  int maxRetries = 2;
+
+  /// An additional test for a failed request's retry eligibility. This function
+  /// will receive the [FinalizedRequest] instance and the [BaseResponse]
+  /// instance from which a decision can be made. If the request should not be
+  /// retried, this should return `false` (or a `Future` that resolves to
```
trentgrover-wf: let's just make the requirement = Future<bool>

lib/src/http/auto_retry.dart
```diff
@@ -0,0 +1,105 @@
...
+
+/// Deciding whether or not to retry a failed request is determined by the
+/// settings defined in fields in this class.
+///
+/// A request will be retried if all of the following are true:
+///
+/// - [enabled] is true
+/// - [forHttpMethods] includes the HTTP method that the failed request used
+/// - [forStatusCodes] includes the response's status code
+/// - [test] (if not null) returns true
```
trentgrover-wf: we should chat through the logic of this set of retry configurations.  it may impose too much configuration for simple use cases.  if all conditions must be met, does that mean you have to ensure that at least 3 of these things are properly configured?

lib/src/http/common/client.dart
```diff
@@ -93,7 +104,11 @@ abstract class CommonClient implements Client {
     if (withCredentials == true) {
       request.withCredentials = true;
     }
-    _requests.add(request);
+    request.autoRetry
+      ..enabled = autoRetry.enabled
+      ..forHttpMethods = autoRetry.forHttpMethods
+      ..forStatusCodes = autoRetry.forStatusCodes
+      ..test = autoRetry.test;
```
maxwellpeterson-wf: Would it make sense to put this logic into an `AutoRetryConfig.clone()` method? Then you could do 

```dart
request.autoRetry = autoRetry.clone;
```

And avoid doing this in more than one place. Would make adding fields easier too, if that ever were to happen.

Just something to consider, this is fine too.

maxwellpeterson-wf:
One note to consider, but either way nice. +1

dustinlessard-wf:
Is there a way to retry with a specific duration between reattempts? Like, wait 500ms between reattempts?

lib/src/http/auto_retry.dart
evanweible-wf: Yeah, I agree :/ Almost all use cases I can think of could be synchronous (async would only be necessary if a streamed response needed to be read), but it is simpler to just require async.

lib/src/http/auto_retry.dart
evanweible-wf: will do

lib/src/http/auto_retry.dart
evanweible-wf: I'm still noodling on this

lib/src/http/common/client.dart
evanweible-wf: I actually don't want to clone the auto retry object between retries - I want the same instance to be shared so that the contextual information (like the current number of attempts) is correct regardless of which request instance you have a reference to.

evanweible-wf:
@dustinlessard-wf I hadn't thought of that, but I like that idea. I'll think about how best to add that (and I may defer it to a follow-up PR).

dustinlessard-wf:
:+1:

evanweible-wf:
@dustinlessard-wf created an issue to implement that later: https://github.com/Workiva/w_transport/issues/102

evanweible-wf:
@trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf I reworked the retry eligibility test logic after some discussion around that area. Also updated the docs for the `AutoRetryConfig` class and added to the readme & changelog.

maxwellpeterson-wf:
+1 - nice

dustinlessard-wf:
+1

lib/src/http/auto_retry.dart
```diff
@@ -0,0 +1,147 @@
...
+
+/// Deciding whether or not to retry a failed request is determined by the
+/// settings defined in fields in this class.
+///
+/// To turn automatic retrying on, set [enabled] to `true`.
+///
+/// The decision to retry a request or not can usually be made based solely on
+/// the request method and the response status code. For this reason, the
+/// [forHttpMethods] and [forStatusCodes] fields define a list of methods and
+/// status codes, respectively, for which it retrying is acceptable.
```
trentgrover-wf: `for which retrying...` (get rid of extra `it`)

trentgrover-wf:
+1 pending minor doc fix

evanweible-wf:
@trentgrover-wf fixed doc string. Once you +1 that commit, I'll rebase and squash.

trentgrover-wf:
+1

dustinlessard-wf:
+1

evanweible-wf:
@trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf rebased and squashed commits.

trentgrover-wf:
+1

maxwellpeterson-wf:
+1

evanweible-wf:
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - verified that examples were still working as expected
- [ ] All unit tests pass

Merging into master.

PR #105, opened 2016-01-05
CP-1243 Add Response.replace() and StreamedResponse.replace()

### LINKED ISSUE #93: Consider an API for easily modifying instances of Response
Currently, to modify an instance of `Response` requires constructing a new `Response` from the original with the desired changes inserted. We would want `Response` instances to remain immutable, but a method for changing certain pieces of the response would be helpful (primary use case being response interceptors). An API similar to `Uri.replace()` would accomplish this.

maxwellpeterson-wf:
## Issue
Fix #93.

This allows the consumer to create new `Response` and `StreamedResponse` objects from an existing instance when they only want to modify one or a few of its fields.

## Changes
**Source:**
- Add new `replace()` method to `Response` and `StreamedResponse`

**Tests:**
- Add test for new `replace()` method on `Response`
- Add test for new `replace()` method on `StreamedResponse`

## Areas of Regression
- None, new API

## Testing
- New tests sufficiently test new functionality
- CI passes

## Code Review
@trentgrover-wf
@evanweible-wf 
@dustinlessard-wf
FYI @jayudey-wf

### DIFF
--- a/lib/src/http/response.dart
+++ b/lib/src/http/response.dart
@@ -54,10 +54,10 @@ abstract class BaseResponse {
     _contentType = http_utils.parseContentTypeFromHeaders(_headers);
   }
 
-  /// Encoding that will be used when decoding the response body. This encoding
-  /// is selected based on [contentType]'s `charset` parameter. If `charset`
-  /// is not given or the encoding name is unrecognized, [LATIN1] is used by
-  /// default ([RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html)).
+  /// Encoding that will be used to decode the response body. This encoding is
+  /// selected based on [contentType]'s `charset` parameter. If `charset` is not
+  /// given or the encoding name is unrecognized, [LATIN1] is used by default
+  /// ([RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html)).
   Encoding get encoding => _encoding;
 
   /// Headers sent with the response to the HTTP request.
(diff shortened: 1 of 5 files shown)


dustinlessard-wf:
Curious what use-case prompted this change?

maxwellpeterson-wf:
@evanweible-wf fixed coverage issue

lib/src/http/response.dart
```diff
@@ -96,6 +101,34 @@ class Response extends BaseResponse {
...
+
+  /// Create a new [Response] using all the values from this instance except
+  /// for the parameters specified.
+  Response replace(
+      {List<int> bodyBytes,
+      String bodyString,
+      int status,
+      String statusText,
+      Map<String, String> headers}) {
+    if (status == null) {
```
jayudey-wf: would you want to use some null aware operators here?

evanweible-wf: it could probably be cleaned up a bit with null aware operators, but w_transport is still compatible (I think) with Dart SDK >=1.9.0 (https://github.com/Workiva/w_transport/blob/master/pubspec.yaml#L40-L41) so unless we really need to use something from a later SDK release or we find that we're not actually compatible that far back, I'd rather not change that.

evanweible-wf:
+1

evanweible-wf:
@maxwellpeterson-wf could you add this to the changelog?

lib/src/http/response.dart
maxwellpeterson-wf: oh good call, wouldnt've thought of that

maxwellpeterson-wf:
@evanweible-wf done

evanweible-wf:
+1 thanks

maxwellpeterson-wf:
@trentgrover-wf @dustinlessard-wf looking for one more reviewer

trentgrover-wf:
+1
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI with tests covering new functionality
- [ ] All unit tests pass

Merging into master.

PR #111, opened 2016-02-04
CP-1330 Add option for SockJS timeout

evanweible-wf:
## Issue
There's currently no way to configure the timeout option for the underlying sockjs-client-dart library. If a socket (using SockJS) takes too long to connect, the connection will be canceled. This is problematic for slower network connections.

## Changes
**Source:**
- Expose the underlying SockJS timeout option via a `sockJSTimeout` optional param.

**Tests:**
- The only way to test this would be an integration test with a custom sockjs server that holds the WS negotiation or maybe the info request until the timeout expires. I'm not sure exactly how to do this, but I can look into if we think it's necessary.

## Areas of Regression
- SockJS socket connection.

## Testing
- CI passes.
- Open the Echo example and check "Use SockJS" along with at least one of "SockJS websocket" and "SockJS xhr-streaming". Should connect and echo as expected.
- Set the SockJS timeout to 1ms - connection should timeout/fail immediately
- Set the SockJS timeout to 1000ms or greater - connection should succeed.

## Code Review
@trentgrover-wf
@maxwellpeterson-wf 
@dustinlessard-wf
@jayudey-wf 

fyi: @bryonmarks-wf @markerickson-wf @stevenosborne-wf

### DIFF
--- a/lib/src/web_socket/browser/sockjs.dart
+++ b/lib/src/web_socket/browser/sockjs.dart
@@ -27,7 +27,8 @@ class SockJSSocket extends CommonWSocket implements WSocket {
   static Future<WSocket> connect(Uri uri,
       {bool debug: false,
       bool noCredentials: false,
-      List<String> protocolsWhitelist}) async {
+      List<String> protocolsWhitelist,
+      Duration timeout}) async {
     if (uri.scheme == 'ws') {
       uri = uri.replace(scheme: 'http');
     } else if (uri.scheme == 'wss') {
@@ -37,7 +38,8 @@ class SockJSSocket extends CommonWSocket implements WSocket {
     sockjs.Client client = new sockjs.Client(uri.toString(),
         debug: debug == true,
         noCredentials: noCredentials == true,
-        protocolsWhitelist: protocolsWhitelist);
+        protocolsWhitelist: protocolsWhitelist,
+        timeout: timeout != null ? timeout.inMilliseconds : null);
 
     // Listen for and store the close event. This will determine whether or
     // not the socket connected successfully, and will also be used later
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@
       ..timeoutThreshold = new Duration(seconds: 10)
       ..autoRetry.enabled = true
       ..autoRetry.forTimeouts = false;
+
+- Added a `Duration sockJSTimeout` config option to `WSocket.connect()`.
+
+    ```dart
+    // browser only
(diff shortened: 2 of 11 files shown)


maxwellpeterson-wf:
+1

trentgrover-wf:
existing `timeoutThreshold` usage uses a `Duration` instead of an `int`.  should we make the `sockJSTimeout` a `Duration` to remain consistent?

evanweible-wf:
@trentgrover-wf depends.. would be inconsistent with the underlying library that uses an int. But I guess we're already abstracting that, so a `Duration` would be fine

trentgrover-wf:
seems like our other libs (like messaging-sdk) are using Durations, so let's try to stick to that

rmconsole3-wf:

lib/src/web_socket/browser/sockjs.dart
```diff
@@ -37,7 +38,8 @@ class SockJSSocket extends CommonWSocket implements WSocket {
     sockjs.Client client = new sockjs.Client(uri.toString(),
         debug: debug == true,
         noCredentials: noCredentials == true,
-        protocolsWhitelist: protocolsWhitelist);
+        protocolsWhitelist: protocolsWhitelist,
+        timeout: timeout.inMilliseconds);
```
trentgrover-wf: tests failing - looks like something related to a null timeout value here so it looks like you need some fallback here

evanweible-wf:
@trentgrover-wf fixed the null timeout.

Also as an FYI, content-shell tests are no longer failing on Travis. Coverage is still broken, but all tests are passing and so is the build so I think we're good now.

trentgrover-wf:
+1

maxwellpeterson-wf:
+1

dustinlessard-wf:
+1, really like that we're using the Duration so much / consistently.

evanweible-wf:
@jayudey-wf ready for merge

evanweible-wf:
@trentgrover-wf @maxwellpeterson-wf @dustinlessard-wf updated example to allow SockJS usage.

dustinlessard-wf:
+1

maxwellpeterson-wf:
+1

trentgrover-wf:
+1

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - tested in example and verified that it was working as expected
- [ ] All unit tests pass 

Merging into master.

PR #133, opened 2016-04-11
cp-1605 : added jitter and maxDuration between retries

dustinlessard-wf:

For consideration:

#### PROBLEM
We don't currently allow jitter with our exponential backoff strategy.

#### SOLUTION
We now allow the optional inclusion of jitter and a maximum retry duration + default maximum retry duration.

Inspired by https://www.awsarchitectureblog.com/2015/03/backoff.html

**Testing Suggestions:**
run unit-tests
run integration tests

**Required:**
@evanweible-wf @maxwellpeterson-wf @trentgrover-wf

**FYI:**
@jayudey-wf


lib/src/http/auto_retry.dart
```diff
@@ -170,18 +171,60 @@ class RetryBackOff {
...
+      }
+    }
+  }
+
   /// Construct a new exponential back-off representation where [duration] is
   /// the base duration from which each delay will be calculated.
-  const RetryBackOff.exponential(this.duration)
-      : method = RetryBackOffMethod.exponential;
+  const RetryBackOff.exponential(this.duration,
+      {bool this.enableJitter, maxDurationInMs})
```
evanweible-wf: Should always use `Duration` - it allows the consumer to decide what unit they want to use. Also `maxDuration` to me isn't very descriptive, duration of what? It's a maximum interval between requests, so maybe `maxInterval` or `maxDelay`.

Also I think I might like `withJitter` as opposed to `enableJitter`.

lib/src/http/auto_retry.dart
```diff
@@ -170,18 +171,60 @@ class RetryBackOff {
...
+      : method = RetryBackOffMethod.exponential,
+        maxDurationInMs = maxDurationInMs ?? defaultMaxDurationInMs;
 
   /// Construct a new fixed back-off representation where [duration] is the
   /// delay between each retry.
-  const RetryBackOff.fixed(this.duration) : method = RetryBackOffMethod.fixed;
+  const RetryBackOff.fixed(this.duration)
```
evanweible-wf: Jitter could be used even with a fixed backoff, couldn't it? Sure it wouldn't help much for a fixed backoff of 1 second, but it could make sense for larger intervals.

lib/src/http/auto_retry.dart
```diff
@@ -170,18 +171,60 @@ class RetryBackOff {
...
   final RetryBackOffMethod method;
 
+  /// Whether to enable jitter or not.
+  final bool enableJitter;
+
+  /// The maximum duration between retries.
+  final maxDurationInMs;
+
+  /// The default maximum duration between retries. (5 minutes)
+  static const defaultMaxDurationInMs = 5 * 60 * 1000;
```
evanweible-wf: Make this and `maxDurationInMs` actual `Duration`s.

lib/src/http/auto_retry.dart
```diff
@@ -170,18 +171,60 @@ class RetryBackOff {
...
+
+  int _baseBackOffFunction(int numAttempts) {
+    return this.duration.inMilliseconds * pow(2, numAttempts);
+  }
+
+  Duration calculate(int numAttempts) {
```
evanweible-wf: I don't think this should be a part of the `RetryBackOff` class. This class is meant to be used by consumers to create representations of the method and details necessary to calculate and implement retry back-off. The consumers shouldn't have access to this calculate method because they shouldn't need to call it. This could instead be moved to the common request implementation.

lib/src/http/auto_retry.dart
```diff
@@ -170,18 +171,60 @@ class RetryBackOff {
...
+    Random random = new Random();
+    if (enableJitter != null && enableJitter == true) {
```
evanweible-wf: this is unnecessary if `enableJitter` is properly initialized via all constructors. The fixed() and none() constructors are fine since they always initialize `enableJitter` to false, the `exponential()` constructor just needs to account for null. I'll make a comment below

lib/src/http/auto_retry.dart
evanweible-wf: If you don't directly assign the optional `enableJitter` param to the `this.enableJitter` above, you can then add an initializer like so below:

```dart
  this.enableJitter = enableJitter == true;
```

lib/src/http/auto_retry.dart
```diff
@@ -170,18 +171,60 @@ class RetryBackOff {
...
+  }
+
   /// Construct a new exponential back-off representation where [duration] is
   /// the base duration from which each delay will be calculated.
-  const RetryBackOff.exponential(this.duration)
-      : method = RetryBackOffMethod.exponential;
+  const RetryBackOff.exponential(this.duration,
+      {bool this.enableJitter, maxDurationInMs})
+      : method = RetryBackOffMethod.exponential,
+        maxDurationInMs = maxDurationInMs ?? defaultMaxDurationInMs;
```
evanweible-wf: So far I've avoided using null-aware operators in w_transport, which allows this lib to support a wider range of SDK versions. A small change here and we can keep that in tact:

```dart
maxDurationInMs = maxDurationInMs != null ? maxDurationInMs : defaultMaxDurationInMs;
```

test/unit/http/auto_retry_test.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright 2015 Workiva Inc.
```
evanweible-wf: With my suggestion to remove `calculate()` from the public API of `RetryBackOff` and into the common request implementation, these tests may not work as well. But I already have a set of integration tests for automatic retrying that gets run against every type of request here: https://github.com/Workiva/w_transport/blob/master/test/integration/http/common_request/suite.dart#L413-L482

I would recommend adding tests there.

evanweible-wf: It would also be fine to move the `calculate()` logic into a utility method that could be tested in isolation - in fact that's probably a good idea. Then you could just use that util method in the common request implementation.

evanweible-wf:
@dustinlessard-wf this only covers 65% of the added functionality - we should get that up to 100%.

lib/src/http/auto_retry.dart
dustinlessard-wf: I think your preference in this case is going to be the x-factor as I don't really have one.  I will point out though that any unit other than millisecond or second is likely to be impractical.  I don't agree that we should always use `Duration` but in this case, I'm indifferent and will change it.  I don't think we should always use it because we don't always want to give consumers the option of using different time units, and it introduces additional overhead where simple integer would suffice.

I too prefer `withJitter`, especially considering the existing code in this repo that uses the same pattern.

lib/src/http/auto_retry.dart
dustinlessard-wf: I'm not familiar with using jitter with a fixed backoff, but will do some research on it.

lib/src/http/auto_retry.dart
evanweible-wf: > any unit other than millisecond or second is likely to be impractical

The fact that there are two options says to me that allowing the use of `Duration` is better than restricting it to MS.

> I don't think we should always use it because we don't always want to give consumers the option of using different time units

It was probably not correct to say we should always use `Duration`, but if there is a limited range of acceptable values, you can enforce that while still giving the consumers the convenience of using `Duration`s.

> it introduces additional overhead where simple integer would suffice.

The `Duration` class is relatively simple and supports a `const` constructor - if the consumer uses a `const` duration, it can be optimized at compile time.

lib/src/http/auto_retry.dart
evanweible-wf: I'm not either, so I'll defer to you here

lib/src/http/auto_retry.dart
trentgrover-wf: :+1:

lib/src/http/auto_retry.dart
```diff
@@ -170,18 +171,60 @@ class RetryBackOff {
...
+        return new Duration(
+            milliseconds: random.nextInt(_baseBackOffFunction(numAttempts)));
+      }
+    } else {
+      if (maxDurationInMs != null) {
+        return new Duration(
+            milliseconds:
+                min(this.maxDurationInMs, _baseBackOffFunction(numAttempts)));
+      } else {
+        return new Duration(milliseconds: _baseBackOffFunction(numAttempts));
```
trentgrover-wf: could be written a bit simpler:
```
  Duration calculate(int numAttempts) {
    int backOffTime = this.duration.inMilliseconds * pow(2, numAttempts); // or _baseBackOffFunction
    if (maxDurationInMs != null) {
      backOffTime = min(this.maxDurationInMs, backOffTime);
    }
    if (enableJitter != null && enableJitter == true) {
      Random random = new Random();
      backOffTime = random.nextInt(backOffTime);
    }
    return new Duration(milliseconds: backOffTime);
  }
```

lib/src/http/auto_retry.dart
dustinlessard-wf: I'm convinced.

dustinlessard-wf: @evanweible-wf , what do you think about renaming RetryBackOff.duration to RetryBackOff.interval to match your suggestion to rename RetryBackOff.maxDuration to maxInterval ?
Since this is a positional parameter, it shouldn't be a breaking change?

lib/src/http/auto_retry.dart
dustinlessard-wf: np, will move.

lib/src/http/auto_retry.dart
```diff
@@ -157,31 +157,50 @@ class RequestAutoRetry extends AutoRetryConfig {
...
+  /// Whether to enable jitter or not.
+  final bool withJitter;
+
+  /// The maximum duration between retries.
+  final Duration maxInterval;
 
-  /// Construct a new fixed back-off representation where [duration] is the
+  /// The default maximum duration between retries. (5 minutes)
+  static const Duration defaultMaxDurationInMs =
+      const Duration(milliseconds: 5 * 60 * 1000);
```
trentgrover-wf:
```
static const Duration defaultMaxInterval = const Duration(minutes: 5);
```

lib/src/http/common/request.dart
```diff
@@ -707,4 +704,60 @@ abstract class CommonRequest extends Object
...
+    switch(autoRetry.backOff.method) {
+      case RetryBackOffMethod.exponential:
+        backOff = _createExponentialBackOff();
+        break;
+      case RetryBackOffMethod.fixed:
+        backOff = _createFixedBackOff();
+        break;
+      case RetryBackOffMethod.none:
+      default:
+        break;
```
trentgrover-wf: does this need to supply a 0 duration instead of null?

lib/src/http/common/request.dart
```diff
@@ -707,4 +704,60 @@ abstract class CommonRequest extends Object
...
+              autoRetry.backOff.maxInterval.inMilliseconds,
+              _baseExponentialBackOffFunction(autoRetry.numAttempts))));
+    } else {
+      backOff = new Duration(
+          milliseconds: min(autoRetry.backOff.maxInterval.inMilliseconds,
+              _baseExponentialBackOffFunction(autoRetry.numAttempts)));
+    }
+
+    return backOff;
+  }
```
trentgrover-wf: replace `_baseExponentialBackOffFunction ` and `_createExponentialBackOff` with something like this?
```
  Duration _createExponentialBackOff() {
    int backOffInMs = autoRetry.backOff.interval.inMilliseconds *
        pow(defaultExponentialMultiplier, autoRetry.numAttempts);
    backOffInMs = min(autoRetry.backOff.maxInterval.inMilliseconds, backOffInMs);

    if (autoRetry.backOff.withJitter != null &&
        autoRetry.backOff.withJitter == true) {
      Random random = new Random();
      backOffInMs = random.nextInt(backOffInMs);
    }
    return new Duration(milliseconds: backOffInMs);
  }
```

lib/src/http/common/request.dart
```diff
@@ -707,4 +704,60 @@ abstract class CommonRequest extends Object
...
+
+  Duration _createFixedBackOff() {
+    Duration backOff;
+    Random random = new Random();
+
+    if (autoRetry.backOff.withJitter != null &&
+        autoRetry.backOff.withJitter == true) {
+      backOff = new Duration(
```
trentgrover-wf: random is only needed in here

lib/src/http/common/request.dart
dustinlessard-wf: It's currently used like this:
```dart
if (backOff != null) {
          await new Future.delayed(backOff);
        }
```

lib/src/http/common/request.dart
```diff
@@ -707,4 +704,51 @@ abstract class CommonRequest extends Object
...
   }
+
+  Duration _createExponentialBackOff() {
+    int backOffInMs = autoRetry.backOff.interval.inMilliseconds *
+        pow(defaultExponentialMultiplier, autoRetry.numAttempts);
+    backOffInMs =
+        min(autoRetry.backOff.maxInterval.inMilliseconds, backOffInMs);
+
+    if (autoRetry.backOff.withJitter != null &&
+        autoRetry.backOff.withJitter == true) {
```
maxwellpeterson-wf: This can be simplified to `if (autoRetry.backOff.withJitter == true) {`

lib/src/http/common/request.dart
```diff
@@ -707,4 +704,51 @@ abstract class CommonRequest extends Object
...
+      backOffInMs = random.nextInt(backOffInMs);
+    }
+    return new Duration(milliseconds: backOffInMs);
+  }
+
+  Duration _createFixedBackOff() {
+    Duration backOff;
+
+    if (autoRetry.backOff.withJitter != null &&
+        autoRetry.backOff.withJitter == true) {
```
maxwellpeterson-wf: This can be simplified to `if (autoRetry.backOff.withJitter == true) {`

lib/src/http/auto_retry.dart
```diff
@@ -177,17 +177,15 @@ class RetryBackOff {
...
 
   /// Construct a new exponential back-off representation where [interval] is
   /// the base duration from which each delay will be calculated.
   const RetryBackOff.exponential(this.interval,
       {bool withJitter, Duration maxInterval})
       : method = RetryBackOffMethod.exponential,
         withJitter = withJitter == true,
-        maxInterval =
-            maxInterval != null ? maxInterval : defaultMaxDurationInMs;
+        maxInterval = maxInterval != null ? maxInterval : defaultMaxInterval;
```
trentgrover-wf: `maxInterval ??= defaultMaxInterval;`?

lib/src/http/auto_retry.dart
```diff
@@ -157,31 +157,48 @@ class RequestAutoRetry extends AutoRetryConfig {
...
+
+  /// Construct a new exponential back-off representation where [interval] is
   /// the base duration from which each delay will be calculated.
-  const RetryBackOff.exponential(this.duration)
-      : method = RetryBackOffMethod.exponential;
+  const RetryBackOff.exponential(this.interval,
+      {bool withJitter, Duration maxInterval})
+      : method = RetryBackOffMethod.exponential,
+        withJitter = withJitter == true,
+        maxInterval = maxInterval != null ? maxInterval : defaultMaxInterval;
```
trentgrover-wf: `maxInterval ??= defaultMaxInterval;`?

trentgrover-wf:
analyzer is failing on the client issue that evan has fixed in the other PR, but there is also 4 failing unit tests

lib/src/http/auto_retry.dart
dustinlessard-wf: I believe @evanweible-wf wanted to support versions of Dart prior to 1.12 (first release with null aware operators).

lib/src/http/common/request.dart
dustinlessard-wf: (indeed)
I can't remember why I avoided that in Dart, but I tested it in Dart 1.12 and it works.

lib/src/http/auto_retry.dart
evanweible-wf: that is correct. It's _probably_ safe to assume most people are on a new enough version of the sdk, but I'd rather make that change all at once and at the next major release.

dustinlessard-wf:
Awaiting updated codecov score, then will update the unit-tests.

evanweible-wf:
@dustinlessard-wf you'll probably have to run coverage locally since it's broken on 1.16 and travis uses 1.16.

trentgrover-wf:
+1
travis ci is failing just because of the known coverage issue

evanweible-wf:
Still need coverage on this:

screen shot 2016-05-26 at 4 05 17 pm

lib/src/http/auto_retry.dart
```diff
@@ -189,9 +189,9 @@ class RetryBackOff {
 
   /// Construct a new fixed back-off representation where [interval] is the
   /// delay between each retry.
-  const RetryBackOff.fixed(this.interval)
+  const RetryBackOff.fixed(this.interval, {bool withJitter})
```
trentgrover-wf: should this have a default value?

trentgrover-wf:
+1, but you need to run dart format :)

dustinlessard-wf:
I needed to force-push some rebased commits, but couldn't force-push to master so created a new pr : https://github.com/Workiva/w_transport/pull/155

PR #136, opened 2016-05-02
CP-1705 Hide dart:html.Client to avoid conflict with w_transport.Client

evanweible-wf:
## Issue
Dart 1.16 exposes a new `Client` class in the `dart:html` library. This conflicts with the `Client` class in w_transport. Thanks for pointing this out @theisensanders-wf.

## Solution
To avoid this, I'm hiding the `Client` class form the `dart:html` import in the one file that this conflict occurs. **This will work for now, but we should probably consider renaming our class since consumers will likely hit this conflict with their own imports.**

## Testing
- [ ] `ddev analyze` runs successfully on Dart 1.16

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf

PR #137, opened 2016-05-02
CP-1709 FormRequest: support fields with multiple values

evanweible-wf:
_Dependent on #136._

## Improvement
Most servers support form requests with multiple-value fields. For example:

```
?foo=bar&foo=baz
```

The above query string would produce this form payload:

```
{ foo: ['bar', 'baz'] }
```

Currently, we don't support this because the `fields` map is of type `Map<String, String>`.

## Changes
- Change the signatures of the `FormRequest.fields` getters and setters to `Map<String, dynamic>`
- Update the utilities for converting from map to query string and vice versa to support fields with multiple values
- Add type checks to verify that each field's value is either `String` or `Iterable<String>`
- Add test to exercise sending a field with multiple values
- Add test to verify that invalid field value type causes an `ArgumentError` to be thrown

## Testing
- [ ] CI passes

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf 

fyi: @brianneal-wf

### DIFF
--- a/lib/src/http/common/form_request.dart
+++ b/lib/src/http/common/form_request.dart
@@ -30,7 +30,7 @@ abstract class CommonFormRequest extends CommonRequest implements FormRequest {
   CommonFormRequest.fromClient(Client wTransportClient, client)
       : super.fromClient(wTransportClient, client);
 
-  Map<String, String> _fields = {};
+  Map<String, dynamic> _fields = {};
 
   @override
   int get contentLength => _encodedQuery.length;
@@ -39,10 +39,10 @@ abstract class CommonFormRequest extends CommonRequest implements FormRequest {
   MediaType get defaultContentType => new MediaType(
       'application', 'x-www-form-urlencoded', {'charset': encoding.name});
 
-  Map<String, String> get fields =>
+  Map<String, dynamic> get fields =>
       isSent ? new Map.unmodifiable(_fields) : _fields;
 
-  set fields(Map<String, String> fields) {
+  set fields(Map<String, dynamic> fields) {
     verifyUnsent();
     if (fields == null) {
       fields = {};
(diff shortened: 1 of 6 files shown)


evanweible-wf:
CI will fail due to coverage issue in Dart 1.16 dart-lang/coverage#107

lib/src/http/common/form_request.dart
```diff
@@ -39,19 +39,26 @@ abstract class CommonFormRequest extends CommonRequest implements FormRequest {
...
     }
     _fields = fields;
   }
 
-  Uint8List get _encodedQuery =>
-      encoding.encode(http_utils.mapToQuery(fields, encoding: encoding));
+  Uint8List get _encodedQuery {
+    fields.forEach((key, value) {
+      if (value is! String && value is! Iterable<String>) {
+        throw new ArgumentError('FormRequest: value of "$key" field must be of '
```
maxwellpeterson-wf: Shouldn't this check go in the setter, not the getter?

evanweible-wf: Ideally, yes this check would be done immediately when adding or modifying fields, but in this case we're exposing a mutable map to allow for incremental additions like this:

```dart
var request = new FormRequest()
  ..fields['foo'] = 'bar'
  ..fields['bar'] = 'baz';
```
This doesn't actually use the setter - that would only happen if the consumer were to set the entire `fields` map. There's no way to catch additions like this. So I placed the check right before we encode the map to a String.

maxwellpeterson-wf: Gotcha

maxwellpeterson-wf:
+1

trentgrover-wf:
+1
@jayudey-wf ready for merge

jayudey-wf:
https://travis-ci.org/Workiva/w_transport/builds/127272420#L4396

QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI exercising new functionatliy
- [ ] All unit tests pass (passed here, https://travis-ci.org/Workiva/w_transport/builds/127272420#L4396)

Failing CI is due to a known coverage issue in Dart 1.16.0 https://github.com/dart-lang/coverage/issues/107

Merging into master.

Issue #134, opened 2016-04-25
WSocket: StreamSubsription onError, onDone aren't honored


stevenosborne-wf:
Consider the following setup: 
```dart
var socket =  await wt.WSocket.connect(
  _natsUri, 
  useSockJS: true, 
  sockJSProtocolsWhitelist: ['websocket', 'xhr-streaming'],
 sockJSTimeout: new Duration(seconds: 1)
);
var streamSub = socket.listen(
  _handleData,
  onError: _handleError,
  onDone: _handleDone
);
```

This will trigger the `_handleData`, `_handleError` and `_handleDone` callbacks appropriately. However, if I reassign the handlers
```dart
streamSub.onData(_newHandleData);
streamSub.onError(_newHandleError);
streamSub.onDone(_newHandleDone);
```
only the `onDone` handler is updated. i.e. lifecycle events and errors are still routed to `_handleDone` and `_handleDone` respectively.

evanweible-wf:
@stevenosborne-wf after looking at this in more depth, the fix is going to require a bit of an overhaul internally with the way we wrap the underlying socket stream and sink. Shouldn't produce any API changes, but it is a bigger effort than I originally thought so I'm going to come back to it in a little while.

PR #142, opened 2016-06-02
CP-1880 Refactor WSocket to fulfill Stream and StreamSink contracts.

evanweible-wf:
_Fixes #134._

## Issue
- `WSocket` extends `Stream` and `StreamSink`, but was not fulfilling those contracts in all scenarios. In particular:
  - After obtaining a `StreamSubscription` instance from a call to `WSocket.listen()`, reassigning the `onData()`, `onError()`, and `onDone()` handlers had no effect.

      ```dart
      var webSocket = await WSocket.connect(...);
      var subscription = webSocket.listen((data) { ... });

      // This does nothing:
      subscription.onData((data) { ... });
      // Same goes for onError() and onDone()
      ```

  - A subscription to a `WSocket` instance did not properly respect pause and resume signals. This could produce a memory leak by buffering WebSocket events indefinitely.

  - A `WSocket` instance was immediately listening to the underlying WebSocket and buffering events from the underlying WebSocket until a listener was registered. This is not how a standard Dart `Stream` works.

  - The SockJS configuration was not properly handling the fact that the SockJS `Client` produces WebSocket events with a broadcast stream.

## Solution

- The `WSocket` implementation has been refactored such that it behaves exactly like a standard Dart `Stream` and `StreamSink`. This involved several things:

  - Proxying the subscription to the underlying WebSocket so that the `onDone()`, `onData()`, and `onError()` handlers could be properly reassigned via the `StreamSubscription` api. This was necessary because the `WSocket` instance is not considered "done" until both the outgoing and the incoming streams are finished closing. Just relying on the "done" event from the subscription to the incoming stream would be inadequate and may fire earlier than desired.

  - Respecting the state of the subscription to the `WSocket` instance. If it is paused, events from the underlying WebSocket are discarded to prevent buffering of events indefinitely which would cause a memory leak. This is in line with how standard Dart streams work - if the subscription is paused, no events should be added to the stream.

  - Handling the fact that the SockJS `Client` produces events from a single broadcast stream. This is different than the native `WebSocket` instances because they are single-subscription streams by default. To account for this, the `WSocket` instance waits to subscribe to the broadcast stream until the `onListen` event. Then if the subscription to the `WSocket` instance is paused, the subscription to the underlying SockJS stream is canceled. Finally, if a paused subscription to the `WSocket` instance is resumed, we re-subscribe to the SockJS stream. This is the recommended solution for dealing with subscriptions to broadcast streams.

- **All of the issues mentioned in the "Issues" section have been addressed, and every `WSocket` instance should now behave exactly as a standard `Stream` and `StreamSink` would, regardless of the platform (VM, browser, SockJS, or mock).**

## Testing
- [ ] CI passes.
- [ ] WebSocket examples still work as expected.

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@sebastianmalysa-wf 
@jayudey-wf 

fyi: @stevenosborne-wf @tayloranderson-wf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,39 @@
 # Changelog
 
+## [2.4.1](https://github.com/Workiva/w_transport/compare/2.4.0...2.4.1)
+_TBD_
+
+- **Bug Fix:** `WSocket` extends `Stream` and `StreamSink`, but was not
+  fulfilling those contracts in all scenarios. In particular:
+
+  - After obtaining a `StreamSubscription` instance from a call to
+    `WSocket.listen()`, reassigning the `onData()`, `onError()`, and `onDone()`
+    handlers had no effect.
+
+      ```dart
+      var webSocket = await WSocket.connect(...);
+      var subscription = webSocket.listen((data) { ... });
+
+      // This does nothing:
+      subscription.onData((data) { ... });
+      // Same goes for onError() and onDone()
+      ```
+
+  - A subscription to a `WSocket` instance did not properly respect pause and
+    resume signals. This could produce a memory leak by buffering WebSocket
+    events indefinitely.
+
+  - A `WSocket` instance was immediately listening to the underlying WebSocket
+    and buffering events from the underlying WebSocket until a listener was
+    registered. This is not how a standard Dart `Stream` works.
+
+  - The SockJS configuration was not properly handling the fact that the SockJS
+    `Client` produces WebSocket events with a broadcast stream.
+
+  - **All of these issues have been addressed, and every `WSocket` instance
+    should now behave exactly as a standard `Stream` and `StreamSink` would,
+    regardless of the platform (VM, browser, SockJS, or mock).**
+
 ## [2.4.0](https://github.com/Workiva/w_transport/compare/2.3.2...2.4.0)
 _May 4, 2016_
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ _TBD_
     should now behave exactly as a standard `Stream` and `StreamSink` would,
     regardless of the platform (VM, browser, SockJS, or mock).**
 
+> The `WSocketCloseEvent` class has been deprecated. This class was only used
+> internally and should not have been exported as a part of the public API.
+
 ## [2.4.0](https://github.com/Workiva/w_transport/compare/2.3.2...2.4.0)
 _May 4, 2016_
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,13 @@ _TBD_
 > The `WSocketCloseEvent` class has been deprecated. This class was only used
 > internally and should not have been exported as a part of the public API.
 
+> The WSocket implementations are no longer registering an `onError` handler for
+> the underlying WebSocket stream. If an error occurs on the server, it will not
+> add the error to the stream, it will just close the connection. As a result,
+> the `MockWSocket.triggerServerError()` method has been deprecated - use
+> `MockWSocket.triggerServerClose()` instead.
+
+
 ## [2.4.0](https://github.com/Workiva/w_transport/compare/2.3.2...2.4.0)
 _May 4, 2016_
 
--- a/lib/src/web_socket/browser/sockjs.dart
+++ b/lib/src/web_socket/browser/sockjs.dart
@@ -20,7 +20,6 @@ import 'package:sockjs_client/sockjs_client.dart' as sockjs;
 
 import 'package:w_transport/src/web_socket/common/w_socket.dart';
 import 'package:w_transport/src/web_socket/w_socket.dart';
-import 'package:w_transport/src/web_socket/w_socket_close_event.dart';
 import 'package:w_transport/src/web_socket/w_socket_exception.dart';
 
 class SockJSSocket extends CommonWSocket implements WSocket {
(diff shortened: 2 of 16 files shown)


lib/src/web_socket/common/w_socket.dart
```diff
@@ -28,74 +28,51 @@ abstract class CommonWSocket extends Stream implements WSocket {
...
-
-  /// The stack trace from the error that was either added to the socket sink or
-  /// received from the socket stream. This will only be populated if an error
-  /// occurs.
+  StreamController _incoming;
+  bool _isIncomingClosed = false;
+  bool _isOutgoingClosed = false;
+  StreamController _outgoing;
   StackTrace _stackTrace;
+  WSocketSubscription _incomingSubscription;
```
evanweible-wf: I'll go back and add some docs for these. I also noticed I removed some docstrings that I didn't mean to, so I'll add those back.

lib/src/web_socket/common/w_socket.dart
```diff
@@ -126,31 +103,58 @@ abstract class CommonWSocket extends Stream implements WSocket {
...
+    // indefinitely.
+    if (!_incoming.isPaused && !_incoming.isClosed) {
+      _incoming.add(data);
+    }
+  }
+
+  void onIncomingDone() {
+    isClosed = true;
+
+    // Now that the socket has closed, capture the close code and reason.
```
evanweible-wf: this comment is outdated

lib/src/web_socket/common/w_socket.dart
```diff
@@ -162,43 +166,39 @@ abstract class CommonWSocket extends Stream implements WSocket {
...
+
+  void onIncomingListen();
+
+  void onIncomingPause();
+
+  void onIncomingResume();
+
+  void onOutgoingData(data);
+
+  void validateOutgoingData(Object data);
```
evanweible-wf: Add some docstrings

lib/src/web_socket/w_socket_subscription.dart
```diff
@@ -0,0 +1,55 @@
+import 'dart:async';
+
+class WSocketSubscription<T> implements StreamSubscription<T> {
```
evanweible-wf: doc string

lib/src/web_socket/w_socket_subscription.dart
```diff
@@ -0,0 +1,55 @@
+import 'dart:async';
```
evanweible-wf: license header

test/integration/ws/common.dart
```diff
@@ -87,133 +89,133 @@ void runCommonWebSocketIntegrationTests({connect(Uri uri), int port}) {
...
+    await new Future.delayed(new Duration(milliseconds: 500));
+    var helper = new WSHelper(webSocket);
     await helper.messagesReceived(3);
 
-    expect(await helper.messages, unorderedEquals(['pong', 'pong', 'pong']));
-    await socket.close();
-  });
+    expect(helper.messages, unorderedEquals(['pong', 'pong', 'pong']));
+    await webSocket.close();
+  }, skip: 'AC changed');
```
evanweible-wf: Remove this test, no longer correct

test/integration/ws/common.dart
```diff
@@ -225,61 +227,169 @@ void runCommonWebSocketIntegrationTests({connect(Uri uri), int port}) {
...
+    webSocket.add('1');
+    webSocket.add('2');
+    // SockJS requires a delay longer than 1 tick for the echos to be received.
+    await new Future.delayed(new Duration(milliseconds: 500));
+
+    subscription.onData(newOnData);
+    webSocket.add('3');
+    webSocket.add('4');
+    // SockJS requires a delay longer than 1 tick for the echos to be received.
+    await new Future.delayed(new Duration(milliseconds: 500));
```
evanweible-wf: It seemed that the SockJS integration tests needed a fairly large delay, but it may have been related to the socket connections not being closed/cleaned up at the end of every test, so I'll try lowering these delays and removing some that may be unnecessary.

tool/dev.dart
```diff
@@ -82,24 +82,6 @@ List<String> _serverOutput;
...
-  await _server.start();
-}
-
-Future _startSockJSServer() async {
-  _sockJSServerOutput = [];
-  _sockJSServer = new TaskProcess('node', ['tool/server/sockjs.js']);
-  _sockJSServer.stdout.listen(_sockJSServerOutput.add);
-  _sockJSServer.stderr.listen(_sockJSServerOutput.add);
-  // todo: wait for server to start
-}
```
evanweible-wf: Buffering the output of the integration servers until the end of the test run was pretty useless - there's no way to map a log entry to the test the generated it, so I switched the test/coverage tasks to use the streaming option so that the logs are placed in line with the tests as they run.

lib/src/web_socket/browser/sockjs.dart
```diff
@@ -61,46 +60,76 @@ class SockJSSocket extends CommonWSocket implements WSocket {
...
+    // the SockJS client are all drawn from a single broadcast stream. To make
+    // it act like a single subscription stream (and thus make it fit the
+    // interface of a standard Stream), we create a subscription when a consumer
+    // listens to this WSocket instance, cancel that subscription when the
+    // consumer's subscription is paused, and re-listen when the consumer
+    // resumes listening. See [onIncomingListen], [onIncomingPause], and
+    // [onIncomingResume].
+
+    // Additional note: the SockJS Client has no error stream, so no need to
+    // listen to for errors.
```
trentgrover-wf: nix `to`?

lib/src/web_socket/browser/sockjs.dart
```diff
@@ -61,46 +60,76 @@ class SockJSSocket extends CommonWSocket implements WSocket {
...
+    // canceling the subscription altogether. When the subscription to this
+    // [WSocket] instance is resumed, we will re-subscribe.
+    webSocketSubscription.cancel();
+  }
+
+  @override
+  void onIncomingResume() {
+    // Resubscribe to the SockJS client's broadcast stream to effectively resume
+    // the consumer's subscription to this [WSocket] instance.
+    webSocketSubscription = _webSocket.onMessage.listen((messageEvent) {
```
trentgrover-wf: is there any risk of orphaning an existing subscription since this doesn't check / cancel any existing subscription before re-assigning?  or is this only used internally and we're confident that it's always used properly?

lib/src/web_socket/common/w_socket.dart
```diff
@@ -28,74 +28,51 @@ abstract class CommonWSocket extends Stream implements WSocket {
...
   CommonWSocket() {
-    // Wait for both the sink and the socket to finish closing before exposing
-    // the close code, close reason, and completing the [_done] completer.
-    // If the socket closed due to an error, the [_done] completer will be
-    // completed with that error.
     _allClosed.future.then((_) {
-      closeCode = _closeEvent.code;
-      closeReason = _closeEvent.reason;
-      _isClosed = true;
+      if (_incomingSubscription?.doneHandler != null) {
```
trentgrover-wf: not a big deal but i think you could replace this with:
```
_incomingSubscription?.doneHandler?.call();
```

lib/src/web_socket/common/w_socket.dart
```diff
@@ -126,31 +103,58 @@ abstract class CommonWSocket extends Stream implements WSocket {
...
     return done;
   }
 
-  void closeSocket(int code, String reason);
+  StreamSubscription listen(void onData(event),
+      {Function onError, void onDone(), bool cancelOnError}) {
+    var sub = _incoming.stream
+        .listen(onData, onError: onError, cancelOnError: cancelOnError);
+    _incomingSubscription = new WSocketSubscription(sub, onDone, onCancel: () {
+      _incomingSubscription = null;
```
trentgrover-wf: does the subscription need to be cancelled here too?

lib/src/web_socket/w_socket_subscription.dart
```diff
@@ -0,0 +1,55 @@
...
+  @override
+  bool get isPaused => _sub.isPaused;
+
+  @override
+  void resume() {
+    _sub.resume();
+  }
+
+  @override
+  void pause([Future resumeSignal]) {
```
trentgrover-wf: should you be doing something with this `resumeSignal`?

test/integration/ws/common.dart
```diff
@@ -225,61 +227,169 @@ void runCommonWebSocketIntegrationTests({connect(Uri uri), int port}) {
...
+    // no subscription yet, messages should be discarded
+    webSocket.add('1');
+    await new Future.delayed(new Duration(milliseconds: 500));
+
+    // setup a subscription, messages should be recorded
+    var sub = webSocket.listen((data) {
+      messages.add(data);
+    });
+    webSocket.add('2');
+    await new Future.delayed(new Duration(milliseconds: 500));
```
trentgrover-wf: do the delays need to be this large to work consistently?  not a huge deal, but time does add up...

trentgrover-wf: maybe a `nextTick` or 2 will be sufficient?

evanweible-wf: agreed, I'm going to try to get these down. It was necessary to get the SockJS suite passing, but that was before I was properly cleaning up all socket connections, so it might be better now

lib/src/web_socket/browser/sockjs.dart
evanweible-wf: The `onIncomingListen`, `onIncomingPause`, `onIncomingResume`, and `onIncomingCancel` callbacks are given to the StreamController upon construction:

```dart
_incoming = new StreamController(
    onListen: onIncomingListen,
    onPause: onIncomingPause,
    onResume: onIncomingResume,
    onCancel: onIncomingCancel);
```

So they're never called manually, they're only called by the `StreamController` implementation. So we can safely assume that `onIncomingResume` will only be called when the controller was previously paused, which would mean we have already canceled the original subscription.

lib/src/web_socket/common/w_socket.dart
evanweible-wf: The code path isn't very clear here, but this `onCancel` callback only gets called when `WSocketSubscription.cancel()` is called, so that would have already happened (https://github.com/Workiva/w_transport/pull/142/files#diff-8abde7635f40c28382f006d30214b7cdR14)

lib/src/web_socket/w_socket_subscription.dart
evanweible-wf: Hmm.. yeah I think I should be

lib/src/web_socket/common/w_socket.dart
evanweible-wf: I shouldn't have used the null-aware operator anyway, I forgot the minimum SDK for this lib is still 1.9.0. I made an issue to bump that minimum version and start using null-awares #144

evanweible-wf:
@trentgrover-wf comments addressed

lib/src/web_socket/common/w_socket.dart
```diff
@@ -28,23 +30,52 @@ abstract class CommonWSocket extends Stream implements WSocket {
   /// no close reason available this property will be `null`.
   String closeReason;
 
+  /// Whether or not this [WSocket] instance is closed or in the process of
+  /// closing.
   bool isClosed = false;
 
+  /// The subscription to the underlying WebSocket (either a browser WebSocket,
+  /// VM WebSocket, SockJS Client, or a mock WebSocket.
```
trentgrover-wf: missing closing `)`

test/integration/ws/common.dart
```diff
@@ -325,31 +329,133 @@ void runCommonWebSocketIntegrationTests(
...
+
+    // resume the subscription, messages should be recorded again
+    c.completeError(new Exception('Ignore. This error is expected.'));
+    await new Future.delayed(new Duration(milliseconds: 100));
+    webSocket.add('3');
+    await new Future.delayed(new Duration(milliseconds: 100));
+
+    expect(messages, orderedEquals(['1', '3']));
+    await webSocket.close();
+  }, skip: 'Can\'t test without the exception causing the test to fail.');
```
trentgrover-wf: test package has a `throwsA` method that you might be able to use for this

test/utils.dart
```diff
@@ -1,5 +1,22 @@
...
+// limitations under the License.
+
 import 'dart:async';
 
-Future nextTick() {
-  return new Future.delayed(new Duration(milliseconds: 1));
+Future nextTick([int numTicks = 1]) async {
+  int i = 0;
+  while (i++ < numTicks) {
+    return new Future.delayed(new Duration(milliseconds: 1));
```
trentgrover-wf: does this work? won't this just return the first future instead of delaying further with a larger `numTick` count?

test/integration/ws/common.dart
evanweible-wf: Yeah unfortunately that only works when the error is thrown synchronously or when a future resolves with an error. In this case, the error is thrown asynchronously and can't be caught like that.

test/utils.dart
evanweible-wf: Nope, it does not. I didn't even mean to commit this change since the integration tests need to wait more than a few ticks. I'll rever this

evanweible-wf:
@trentgrover-wf fixed. Also forgot to mention - I tried using nextTick() in the integration tests, but the SockJS suite still fails. 100ms was about as low as I could get the waits while still having the tests pass consistently.

trentgrover-wf:
+1

CHANGELOG.md
```diff
@@ -1,5 +1,42 @@
 # Changelog
 
+## [2.4.1](https://github.com/Workiva/w_transport/compare/2.4.0...2.4.1)
```
dustinlessard-wf: These release notes are insanely good!

dustinlessard-wf:
+1

trentgrover-wf:
+1

dustinlessard-wf:
+1

sebastianmalysa-wf:
+1

sebastianmalysa-wf:
+1

stevenosborne-wf:
+10 Thanks Evan!

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - performed and documented by Steven
- [ ] All unit tests pass 

Merging into master.

PR #155, opened 2016-06-20
Cp 1605 : added jitter and maxDuration between retries

dustinlessard-wf:

This is a continuation of https://github.com/Workiva/w_transport/pull/133
For consideration:

#### PROBLEM
We don't currently allow jitter with our exponential backoff strategy.

#### SOLUTION
We now allow the optional inclusion of jitter and a maximum retry duration + default maximum retry duration.

Inspired by https://www.awsarchitectureblog.com/2015/03/backoff.html

**Testing Suggestions:**
run unit-tests
run integration tests

**Required:**
@evanweible-wf @maxwellpeterson-wf @trentgrover-wf

**FYI:**
@jayudey-wf


lib/src/http/common/request.dart
```diff
@@ -707,4 +704,49 @@ abstract class CommonRequest extends Object
...
+    switch (autoRetry.backOff.method) {
+      case RetryBackOffMethod.exponential:
+        backOff = _createExponentialBackOff();
+        break;
+      case RetryBackOffMethod.fixed:
+        backOff = _createFixedBackOff();
+        break;
+      case RetryBackOffMethod.none:
+      default:
+        break;
```
evanweible-wf: This should probably throw, wouldn't want to return null here.

evanweible-wf: Nevermind, there's a null-check above

lib/src/http/common/request.dart
```diff
@@ -707,4 +704,49 @@ abstract class CommonRequest extends Object
...
+        min(autoRetry.backOff.maxInterval.inMilliseconds, backOffInMs);
+
+    if (autoRetry.backOff.withJitter == true) {
+      Random random = new Random();
+      backOffInMs = random.nextInt(backOffInMs);
+    }
+    return new Duration(milliseconds: backOffInMs);
+  }
+
+  Duration _createFixedBackOff() {
```
evanweible-wf: Could you move this method and the one above out into a utility file? There's no hard dependency on this class, they could be updated to accept `interval`, `maxInterval`, and `withJitter` as parameters. Then you can unit test this math without needing to deal with requests.

evanweible-wf:
Still missing some coverage here:

screen shot 2016-06-20 at 8 31 25 am

lib/src/http/common/request.dart
dustinlessard-wf: Will do.

dustinlessard-wf:
@evanweible-wf , please take a look.  I moved the code into a utility class, added more extensive unit-tests and noticed that despite my belief that we've unit-tested all of the code in the backoff.dart file, it doesn't seem to be represented in the code coverage.

test/unit/http/common/backoff_test.dart
```diff
@@ -0,0 +1,184 @@
...
+
+import 'package:w_transport/src/http/auto_retry.dart';
+
+import '../../../naming.dart';
+import 'package:w_transport/src/http/requests.dart';
+import 'package:w_transport/src/http/common/backoff.dart';
+import 'package:w_transport/w_transport_mock.dart';
+import 'dart:math';
+
+reqFactory({bool withBody: false}) {
```
evanweible-wf: This doesn't appear to be used anywhere

test/unit/http/common/backoff_test.dart
```diff
@@ -0,0 +1,184 @@
...
+  if (!withBody) return new Request();
+  return new Request()..body = 'body';
+}
+
+void main() {
```
evanweible-wf: You'll have to import this file and execute this main block in the platform independent test suite: https://github.com/Workiva/w_transport/blob/master/test/unit/platform_independent_unit_test.dart

test/unit/http/common/backoff_test.dart
```diff
@@ -0,0 +1,184 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@TestOn('browser')
+import 'dart:async';
+import 'dart:html';
```
evanweible-wf: I don't think this has any dependency on `dart:html`, does it? If not, should remove this import and change the `@TestOn` annotation to `@TestOn('vm || browser');`

test/unit/http/common/backoff_test.dart
```diff
@@ -0,0 +1,179 @@
...
+
+import 'package:w_transport/src/http/auto_retry.dart';
+
+import '../../../naming.dart';
+import 'package:w_transport/src/http/requests.dart';
+import 'package:w_transport/src/http/common/backoff.dart';
+import 'package:w_transport/w_transport_mock.dart';
+import 'dart:math';
+
+void main() {
```
evanweible-wf: One of my comments got buried with the last commit, so I'll repost:

You'll have to import this file and execute this main block in the platform independent test suite: https://github.com/Workiva/w_transport/blob/master/test/unit/platform_independent_unit_test.dart

evanweible-wf:
@dustinlessard-wf looks like there's one failing test.

test/unit/http/request_test.dart
```diff
@@ -1037,6 +1039,57 @@ _runAutoRetryTestSuiteFor(
...
+          ..backOff = new RetryBackOff.exponential(
+              new Duration(milliseconds: 25),
+              withJitter: true);
+        request.get(uri: requestUri);
+
+        // 1st attempt = immediate
+        // 2nd attempt = +0 to 50s (25*2^1)
+        // 3rd attempt = +0 to 100s (25*2^2) + 2nd attempt
+        // 4th attempt = +0 to 200s (25*2^3) + 3rd attempt
+        await new Future.delayed(new Duration(milliseconds: 350));
```
evanweible-wf: Could you add another expectation just to assert that they're not all happening immediately? Something like

```dart
await new Future.delayed(new Duration(milliseconds: 50));
expect(request.autoRetry.numAttempts, greaterThan(0));
expect(request.autoRetry.numAttempts, lessThan(4));
await new Future.delayed(new Duration(milliseconds: 300));
expect(request.autoRetry.numAttempts, equals(4));
```

evanweible-wf: Talked with Dustin about this, not going to worry about it.

evanweible-wf:
@dustinlessard-wf this looks good, could you rebase against the latest and squash into 1 commit?

dustinlessard-wf:
Rebased here : https://github.com/Workiva/w_transport/pull/158

Issue #159, opened 2016-06-22
Handle request with no encoding and no charset


evanweible-wf:
```
dart:core                                                          Object.noSuchMethod
package:w_transport/src/http/response_format_exception.dart 55:38  ResponseFormatException.message
package:w_transport/src/http/response_format_exception.dart 62:50  ResponseFormatException.toString
dart:core                                                          _StringBase._interpolate
package:w_transport/src/http/request_exception.dart 68:28          RequestException.message
package:frugal/src/transport/f_http_transport.dart 96:67           FHttpClientTransport.flush.<async>
dart:async                                                         _Completer.completeError
package:w_transport/src/http/common/request.dart 721:1             CommonRequest._send.<async>
===== asynchronous gap ===========================
dart:async                                                         _Completer.completeError
package:frugal/src/transport/f_http_transport.dart 126:1           FHttpClientTransport.flush.<async>
dart:async                                                         _Completer.completeError
package:w_transport/src/http/common/request.dart 721:1             CommonRequest._send.<async>
===== asynchronous gap ===========================
dart:async                                                         _asyncErrorWrapperHelper
package:frugal/src/transport/f_http_transport.dart                 FHttpClientTransport.flush
test/transport/f_http_transport_test.dart 70:23                    main.<fn>.<fn>.<async>
===== asynchronous gap ===========================
dart:async                                                         Future.Future.microtask
test/transport/f_http_transport_test.dart                          main.<fn>.<fn>
The null object does not have a getter 'name'.
NoSuchMethodError: method not found: 'name'
Receiver: null
Arguments: []
```

Looking at the stack trace above, it first looks like it's an issue decoding the response body, but I believe the `ResponseFormatException` is actually masking a different bug. In the `HttpBody` implementation,
we are catching all exceptions when trying to encode or decode the response body:

https://github.com/Workiva/w_transport/blob/master/lib/src/http/http_body.dart#L87-L91
https://github.com/Workiva/w_transport/blob/master/lib/src/http/http_body.dart#L100-L104

1. That should be changed - we should only catch an `ArgumentError` for encode and `FormatException` for decode.
2. I believe the real exception is that `encoding` is null, because the stack trace points to the line where the `ResponseFormatException` class is trying to print the name of `encoding`: https://github.com/Workiva/w_transport/blob/master/lib/src/http/response_format_exception.dart#L55 

[...]

PR #161, opened 2016-06-22
CP-1966 Handle request with no encoding and no charset

evanweible-wf:
_Fixes #159._

## Issue
#159 reported an issue when setting a request's content-type manually. If that content-type has no `charset` parameter, and the request's encoding was not set, you will hit an RTE in `HttpBody` when trying to read the body and a conversion is required (bytes->string or vice versa). This is because the `HttpBody` constructors have an optional `fallbackEncoding` parameter that has no default value. For responses, we pass in a fallback encoding of `LATIN1` per RFC 2616, but we aren't passing in a fallback encoding at all for request bodies.

This issue hadn't manifested itself before because most requests set the content-type automatically, including the charset, but #153 introduced the ability to manually set the content-type.

## Solution
- Update `HttpBody.fromString()` and `HttpBody.fromBytes()` to give the optional `fallbackEncoding` parameter a default value of `UTF8`.
- Add a null check that will throw an `ArgumentError` if `fallbackEncoding` happens to be null
- Additionally, a null check for encoding was added before trying to read the name in `ResponseFormatException`.
- Bonus: When attempting to decode/encode a request/response body, we now only catch the expected error instead of any exception (ArgumentError for encode, FormatException for decode). This will prevent us from swallowing unexpected exceptions erroneously wrapping them in a `ResponseFormatException`, which is what happened in the stack trace from #159.

## Testing
- [ ] CI passes
- [ ] HTTP examples work as expected
- [ ] @stevenosborne-wf's tests are fixed

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf 
@sebastianmalysa-wf 
@srinivasdhanwada-wf 

fyi: @stevenosborne-wf

### DIFF
--- a/lib/src/http/http_body.dart
+++ b/lib/src/http/http_body.dart
@@ -56,10 +56,31 @@ class HttpBody extends BaseHttpBody {
   Encoding _encoding;
 
   /// Construct the body to an HTTP request or an HTTP response from bytes.
+  ///
+  /// If [encoding] is given, it will be used to encode/decode the body.
+  ///
+  /// Otherwise, the `charset` parameter from [contentType] will be mapped to an
+  /// Encoding supported by Dart.
+  ///
+  /// If a `charset` parameter is not available or the value is unrecognized,
+  /// [fallbackEncoding] will be used.
+  ///
+  /// If [fallbackEncoding] is `null`, UTF8 will be the default encoding.
+  ///
+  /// If an encoding cannot be parsed from the content-type header (via the
+  /// `charset` param), then [fallbackEncoding] will be used (UTF8 by default).
   HttpBody.fromBytes(MediaType this.contentType, List<int> bytes,
-      {Encoding fallbackEncoding}) {
-    _encoding = http_utils.parseEncodingFromContentType(contentType,
-        fallback: fallbackEncoding);
+      {Encoding encoding, Encoding fallbackEncoding}) {
+    if (fallbackEncoding == null) {
+      fallbackEncoding = UTF8;
+    }
+    if (encoding != null) {
+      _encoding = encoding;
+    } else {
+      _encoding = http_utils.parseEncodingFromContentType(contentType,
+          fallback: fallbackEncoding);
+    }
+
     if (bytes == null) {
       bytes = [];
     }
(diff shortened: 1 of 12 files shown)


lib/src/http/http_body.dart
```diff
@@ -56,8 +56,14 @@ class HttpBody extends BaseHttpBody {
...
 
   /// Construct the body to an HTTP request or an HTTP response from bytes.
+  ///
+  /// If an encoding cannot be parsed from the content-type header (via the
+  /// `charset` param), then [fallbackEncoding] will be used (UTF8 by default).
   HttpBody.fromBytes(MediaType this.contentType, List<int> bytes,
-      {Encoding fallbackEncoding}) {
+      {Encoding fallbackEncoding: UTF8}) {
+    if (fallbackEncoding == null) {
+      throw new ArgumentError.notNull('fallbackEncoding');
```
maxwellpeterson-wf: Would it be okay to set it to `UTF8` if it is null here instead of throwing an error?

evanweible-wf:
Sorry, hold of on this briefly, just thought of one improvement.

lib/src/http/http_body.dart
evanweible-wf: Yeah that's better.

evanweible-wf:
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf 
@sebastianmalysa-wf 
@srinivasdhanwada-wf 

Ready to review now.

maxwellpeterson-wf:
+1

It looks like there's one branch that isn't covered... specifying the optional `encoding` to `HttpBody.fromBytes()`.

evanweible-wf:
@maxwellpeterson-wf good catch, my new HttpBody tests were supposed to test both constructors but I copy pasted and didn't change `fromString` to `fromBytes`. Should be fixed now.

maxwellpeterson-wf:
+1

sebastianmalysa-wf:
+1

lib/src/http/http_body.dart
```diff
@@ -56,21 +56,63 @@ class HttpBody extends BaseHttpBody {
...
+  ///
+  /// If [fallbackEncoding] is `null`, UTF8 will be the default encoding.
+  ///
+  /// If an encoding cannot be parsed from the content-type header (via the
+  /// `charset` param), then [fallbackEncoding] will be used (UTF8 by default).
   HttpBody.fromString(MediaType this.contentType, String body,
-      {Encoding fallbackEncoding}) {
-    _encoding = http_utils.parseEncodingFromContentType(contentType,
-        fallback: fallbackEncoding);
+      {Encoding encoding, Encoding fallbackEncoding: UTF8}) {
```
trentgrover-wf: there's some inconsistency in this `fallbackEncoding` code.  why does this method signature have a default value but the other does not?  if there's a default value, no need to the null check to set UTF8

evanweible-wf: Good catch, I had it this way at first but switched to Max's suggestion of just populating `fallbackEncoding` to UTF8 if it's null instead of throwing an ArgumentError. I'll fix this

evanweible-wf: Fixed

evanweible-wf:
@trentgrover-wf @maxwellpeterson-wf @sebastianmalysa-wf 

Addressed Trent's comment.

maxwellpeterson-wf:
+1

trentgrover-wf:
+1
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI with tests utilizing new functionality
- [ ] All unit tests pass

Merging into master.

PR #176, opened 2016-07-14
CP-2059 Support auto-retying requests that fail with a null response

### LINKED ISSUE #173: Allow HTTP requests with status 0 to be retried
When an XHR request is made from the browser, there are some special case failures that may occur. The browser may not be able to complete the request because there is no network connection, or it may refuse to complete the request as a security measure (example: server did not return access-control headers, thus rejecting the cross-origin request). In the network tab, you'll likely see that the request returned with a status code of 0. This is not an official HTTP status code, but rather a special code used by browsers to indicate the type of failure to the developer.

In code, a response will most likely not be received at all. Instead, an error event will be triggered on the XHR instance. With our current implementation, this results in the error being raised (and then wrapped in a `RequestException`) and a null `response`. Additionally, our "can retry" logic treats a null `response` as a non-retryable failure. I'm not sure this should be the case.

On one hand, we don't know exactly why a request failed when it fails like this, so it may be useless to retry it (CORS). On the other hand, one of the more common scenarios in which this would happen is a lack of network connection (offline mode or intermittent connectivity), which would benefit from retrying.

**tl;dr:** we should consider allowing retries for failed requests where the failure is unknown.

maxwellpeterson-wf:
## Issue
Fix #173 

## Solution
This will allow auto-retrying of requests that have null responses. Consumers will need to add a `test` method to `request.autoRetry` that checks for a `null` response and returns `true` to enable the functionality.

@evanweible-wf thoughts?

## Testing
Added an integration test that causes a `null` response and ensures it gets retried.

@trentgrover-wf
@evanweible-wf
@dustinlessard-wf
@jayudey-wf
@sebastianmalysa-wf 
@srinivasdhanwada-wf

### DIFF
--- a/lib/src/http/common/request.dart
+++ b/lib/src/http/common/request.dart
@@ -492,13 +492,17 @@ abstract class CommonRequest extends Object
 
     // If the request failed due to exceeding the timeout threshold, check if
     // it is configured to retry for timeouts.
-    if (requestException.error is TimeoutException && autoRetry.forTimeouts)
-      return true;
+    if (requestException.error is TimeoutException) {
+      return autoRetry.forTimeouts;
+    }
 
-    if (response == null) return false;
-
-    bool willRetry = autoRetry.forHttpMethods.contains(method) &&
-        autoRetry.forStatusCodes.contains(response.status);
+    bool willRetry = autoRetry.forHttpMethods.contains(method);
+    if (response != null && response.status != null) {
+      willRetry =
+          willRetry && autoRetry.forStatusCodes.contains(response.status);
+    } else {
+      willRetry = false;
+    }
     if (autoRetry.test != null) {
       willRetry = await autoRetry.test(request, response, willRetry);
     }
--- a/test/integration/http/common_request/browser_test.dart
+++ b/test/integration/http/common_request/browser_test.dart
@@ -16,8 +16,10 @@
 library w_transport.test.integration.http.common_request.browser_test;
 
 import 'package:test/test.dart';
+import 'package:w_transport/w_transport.dart';
 import 'package:w_transport/w_transport_browser.dart';
 
+import '../../integration_paths.dart';
 import '../../../naming.dart';
 import 'suite.dart';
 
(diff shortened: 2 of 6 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

rmconsole-wf:
@maxwellpeterson-wf
* It looks like you have not included any tickets in the title of this PR, but this repo requires tickets for all pull requests. Please create a ticket for this change, or click <a href='https://w-rmconsole.appspot.com/api/v1/pulls/77526396/addTicketToTitle'>here</a> to have Rosie create one for you.

lib/src/http/common/request.dart
```diff
@@ -495,10 +495,8 @@ abstract class CommonRequest extends Object
     if (requestException.error is TimeoutException && autoRetry.forTimeouts)
       return true;
 
-    if (response == null) return false;
-
     bool willRetry = autoRetry.forHttpMethods.contains(method) &&
-        autoRetry.forStatusCodes.contains(response.status);
+        autoRetry.forStatusCodes.contains(response?.status ?? 0);
```
evanweible-wf: Could you do this without null-aware operators for now?

maxwellpeterson-wf: Fixed

lib/src/http/common/request.dart
```diff
@@ -495,10 +495,14 @@ abstract class CommonRequest extends Object
     if (requestException.error is TimeoutException && autoRetry.forTimeouts)
       return true;
 
-    if (response == null) return false;
+    // TODO replace with null-aware operators once this library uses them.
+    var status = response != null ? response.status : 0;
+    if (status == null) {
```
trentgrover-wf: status can never be null here, right?

trentgrover-wf: nm, i see...

trentgrover-wf:
I'm not sure all of this is necessary.  In dart it's permissible for a List<int> to contain null as an entry, so consumers can specify `null` in `forStatusCodes` and it will work (and it would be more self documenting)

maybe you only need:
```
var status = response != null ? response.status : null;
```

maxwellpeterson-wf:
`0` is a real status code used by browsers though for a failed request, so this is more in line with that.

@evanweible-wf thoughts?

evanweible-wf:
I changed my mind on the 0 status code - since it's not actually in the HTTP spec, we shouldn't introduce any knowledge of it even if browsers do use it. I don't like having consumers put `null` in their `forStatusCodes` list, either. But they can still detect it by supplying a `test` method that checks for a `null` response, so I think we should go that route.

maxwellpeterson-wf:
Done

trentgrover-wf:
+1

sebastianmalysa-wf:
+1

evanweible-wf:
+1

evanweible-wf:
+1

trentgrover-wf:
+1

trentgrover-wf:
+1

evanweible-wf:
+1

trentgrover-wf:
+1

maxwellpeterson-wf:
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI with test exercising functionality
- [ ] All unit tests pass

Merging into master.

Issue #140, opened 2016-05-24
CP-2243 Rename `Client`


evanweible-wf:
Dart 1.16 introduced a `Client` class to the `dart:html` import, which conflicts with the `Client` class in this library. It can be worked around by importing either `dart:html` or `package:w_transport/w_transport.dart` with a prefix, or by hiding `Client` from `dart:html` if it is unneeded. However, this is likely to create some confusion and is less than an ideal solution.

We should rename `Client` to something more explicit. We should keep in mind the potential addition of an HTTP/2 client as well.

This is obviously a breaking change, so the first step will be to introduce a copy of the `Client` class under the new name and deprecate `Client` class. This will happen in 3.0.0. Then in 4.0.0 we can remove `Client`.

evanweible-wf:
Merged to 3.0.0-wip in #209

PR #209, opened 2016-08-16
CP-2243 Rename Client to HttpClient

evanweible-wf:
_Fixes #140._

## Problem
A `Client` class was recently added to the `dart:html` library, which conflicts with the `Client` class in `w_transport`. This can be worked around, but is annoying.

## Solution
Deprecate `Client` and add a new `HttpClient` class to replace it. They are identical in functionality. All public APIs still use the deprecated `Client` class to maintain backwards compatibility. `Client` and all references will be removed completely in 4.0.0.

## Changes
- Deprecate `Client`
- Add `HttpClient`

## Documentation
Docs/changelog will be updated in my docs overhaul WIP branch.

## Testing
- [ ] CI passes

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf 
@sebastianmalysa-wf 
@srinivasdhanwada-wf

### DIFF
--- a/lib/src/http/http_client.dart
+++ b/lib/src/http/http_client.dart
@@ -12,11 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import 'package:w_transport/src/http/mock/http_client.dart';
-import 'package:w_transport/src/http/http_client.dart';
+import 'package:w_transport/src/http/client.dart';
+import 'package:w_transport/src/platform_adapter.dart';
 
-/// A mock implementation of an HTTP client. Factory methods simply return the
-/// mock implementations of each request. Since the mock request implementations
-/// don't ever actually send an HTTP request, this client doesn't need to do
-/// anything else.
-class MockClient extends MockHttpClient implements HttpClient {}
+/// An HTTP client acts as a single point from which many requests can be
+/// constructed. All requests constructed from a client will inherit [headers],
+/// the [withCredentials] flag, and the [timeoutThreshold].
+///
+/// On the server, the Dart VM will also be able to take advantage of cached
+/// network connections between requests that share a client.
+abstract class HttpClient extends Client {
+  factory HttpClient() => PlatformAdapter.retrieve().newHttpClient();
+}
--- a/lib/src/browser_adapter.dart
+++ b/lib/src/browser_adapter.dart
@@ -14,9 +14,9 @@
 
 import 'dart:async';
 
-import 'package:w_transport/src/http/browser/client.dart';
+import 'package:w_transport/src/http/browser/http_client.dart';
 import 'package:w_transport/src/http/browser/requests.dart';
-import 'package:w_transport/src/http/client.dart';
+import 'package:w_transport/src/http/http_client.dart';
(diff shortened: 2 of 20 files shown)


aviary2-wf:
## Raven
Number of Findings: 3
* Watched keyword ```withCredentials``` in ```lib/src/http/common/client.dart``` line(s) ['47', '94', '102-103'] removed
* Watched keyword ```withCredentials``` in ```lib/src/http/http_transport_client.dart``` line(s) ['20'] added
* Watched keyword ```withCredentials``` in ```lib/src/http/common/http_transport_client.dart``` line(s) ['47', '94', '102-103'] added

sebastianmalysa-wf:
+1

maxwellpeterson-wf:
After thinking about this more I don't love the change.

`HttpTransportClient` is still as generic as `Client`. By that I mean Dart could easily include a new class in the next sdk version that is called `HttpTransportClient`. I'm not betting that they will, but I'm wondering if it is worth a breaking change that moves to a significantly longer name that isn't even unique to this library. The new name doesn't give you any clue as to which library it is from. There's really no benefit other than preventing the need to namespace your import or use `hide` when `dart:html` is also imported.

Should we consider `WHttpClient`? 

Or, perhaps we scrap this idea altogether and allow people to namespace their imports or use `hide` on the `dart:html` import. We can't expect that every class in every package we make is going to be unique to that package. This is especially true with a library like this that lives so close to the browser's api. What if Dart or other prominent package adds classes called `Request`, `JsonRequest`, `MultipartRequest`,`Http`, `HttpBody`, `RequestException`, `Response`, etc? Will we rename those also?

dustinlessard-wf:
Hmm...
My feeling on the topic is that `HttpTransportClient` is not as generic as `Client`, but am in total agreement that `Request`, `Response` and many other classes are exceedingly generic and could easily suffer from the same issue.

One consideration is how w_transport sits at the base of many of our current / future products.  Do we risk the pain of future conflicts or accept the pain now and rename the likeliest of offenders.

Whatever we decide, we should consider how it may become the pattern that others follow.  Personally, I like the idea of choosing unique names in an effort to avoid future naming collisions.  I think our consumers will thank us.

trentgrover-wf:
naming consensus reached in separate conversation:
- rename Client -> HttpClient
- rename WSocket -> WebSocket
- recommend importing with namespace to avoid collisions with dart sdk bits

maxwellpeterson-wf:
+1

lib/src/http/http_client.dart
```diff
@@ -0,0 +1,26 @@
+// Copyright 2015 Workiva Inc.
```
trentgrover-wf: guess we should update copyright date at some point...

trentgrover-wf:
+1
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI
- [ ] All unit tests pass 

Merging into WIP.

chadknight-wf:
+1 security

evanweible-wf:

PR #220, opened 2016-09-09
CP-2657 Standardize variable declarations

evanweible-wf:
## Improvement
Now that strong-mode analyzer and linter have been enabled, I wanted to standardize the way variables are declared in this codebase. The goal was to provide strong typing (which is guaranteed by the strong-mode analysis) while avoiding overly-verbose syntax (like specifying the type on both the left- and right-hand side everywhere.

## Changes
In this process, I developed a set of guidelines based on the "Effective Dart" guides as well as input from the Dart community that I've come across.

1. **Always type public members.**

  This one is pretty straightforward and there's even a linter rule for it. Anything exposed publicly should be strongly typed so that there's no confusion for consumers.

1. **If the return type is unknown, but you expect it to be an object (in other words, you expect it to implement `toString()` and `hashCode`): use `Object`.**

  There are times when the return type just isn't known. This is totally fine, but it should be communicated explicitly by using `Object` instead of omitting the type.

1. **If the return type is _truly unknown_: use `dynamic`.**

  If the return type is unknown _and_ you can't guarantee that it will be an `Object`, use `dynamic`.

1. **Don't use `var`.**

  This one is probably the most surprising because of how often it is used. [Jim Beveridge originally pointed me in this direction](https://github.com/Workiva/w_flux/pull/63#r67577219) and I've convinced myself that it's a good idea.

  Using `var` tells the analyzer absolutely nothing. Using `final` tells the analyzer that the variable should not be overwritten.

  If the value doesn't change, then using `final` instead of `var` is an easy switch that tells the analyzer more without requiring any additional changes.

    ```dart
    // This variable can be reassigned later to any type.
    var number = 4;

    // This variable cannot be reassigned.
    // We have a guarantee of the value and type.
    final number = 4;
    ```

  If the value can change after the initial declaration/assignment, then we should leverage typing so that the analyzer can ensure we don't set the variable to an incompatible type when it does change.

    ```dart
    // If we use `var`, we have no guarantee on the type of `result`.
    // Theoretically, `result` could be assigned to an incompatible type later.
    var result;
    if (condition) {
      result = 'success';
    } else {
      result = 'failure';
    }
    return result;

    // If we specify the expected type, we regain the type guarantee and
    // the analyzer can help us detect invalid assignments/operations.
    String result;
    if (condition) {
      result = 'success';
    } else {
      result = 'failure';
    }
    return result;
    ```

1. **Omit the type on the LHS if it can be sufficiently inferred from the RHS.**

  This applies only to non-public members because of rule 1. The goal of this rule is to avoid unnecessarily verbose typing like the following:

    ```dart
    StreamController<List<int>> byteStreamController = new StreamController<List<int>>();
    ```

  If the RHS is fully-typed, we can omit the typing on the LHS while still satisfying the strong-mode analysis:

    ```dart
    final byteStreamController = new StreamController<List<int>>();
    ```

1. **Don't omit the generic type. If no return value is expected, use `Null`.**

    ```diff
    - Future
    + Future<Null>
    - Completer
    + Completer<Null>
    ```

## Testing
- [ ] CI passes.

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf 
@sebastianmalysa-wf

### DIFF
--- a/example/common/global_example_menu_component.dart
+++ b/example/common/global_example_menu_component.dart
@@ -18,24 +18,22 @@ import 'dart:html';
 import 'package:react/react.dart' as react;
 import 'package:w_transport/w_transport.dart';
 
-const int _pollingInterval = 10; // 10 seconds
-
 void renderGlobalExampleMenu({bool nav: true, bool serverStatus: false}) {
   // Insert a container div within which we will mount the global example menu.
-  var container = document.createElement('div');
+  final container = document.createElement('div');
   container.id = 'global-example-menu';
   document.body.insertBefore(container, document.body.firstChild);
 
   // Use react to render the menu.
-  var menu =
+  final menu =
       globalExampleMenuComponent({'nav': nav, 'serverStatus': serverStatus});
   react.render(menu, container);
 }
 
 Future<bool> _ping(Uri uri) async {
   try {
-    Response response = await Http.get(uri);
-    return response.status == 200;
+    await Http.get(uri);
+    return true;
   } on RequestException {
     return false;
   }
@@ -63,12 +61,12 @@ class GlobalExampleMenuComponent extends react.Component {
   @override
   void componentWillMount() {
     if (this.props['serverStatus']) {
-      _pingServer().then((bool status) {
+      _pingServer().then((status) {
         this.setState({'serverOnline': status});
       });
       serverPolling =
           new Timer.periodic(new Duration(seconds: 4), (Timer timer) async {
-        bool status = await _pingServer();
+        final status = await _pingServer();
         this.setState({'serverOnline': status});
       });
     }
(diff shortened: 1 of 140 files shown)


aviary2-wf:
## Raven
Number of Findings: 2
* Watched keyword ```withCredentials``` in ```lib/src/http/finalized_request.dart``` line(s) ['39'] added
* Watched keyword ```withCredentials``` in ```lib/src/http/finalized_request.dart``` line(s) ['43'] removed

rmconsole-wf:
@evanweible-wf
* It looks like you have not included any tickets in the title of this PR, but this repo requires tickets for all pull requests. Please create a ticket for this change, or click <a href='https://w-rmconsole.appspot.com/api/v1/pulls/84705975/addTicketToTitle'>here</a> to have Rosie create one for you.

rmconsole-wf:

maxwellpeterson-wf:
+1

dustinlessard-wf:
+1 - great guidelines!

trentgrover-wf:
+1 to all those guidelines

trentgrover-wf:
@jayudey-wf ready for merge

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI
- [ ] All unit tests pass

Merging into master.

evanweible-wf:

PR #222, opened 2016-09-19
CP-2658 Refactor platform configuration

evanweible-wf:
_Fixes #125._

## Improvement
Currently, configuring w_transport for a platform (browser, VM, test) is all-or-nothing. This results in some less-than-ideal consequences:

- The platform can only be configured globally. In other words, the configuration is applied to every request and every WebSocket.

- Libraries cannot specify which platform they'd like to use.

  > If a library were to configure w_transport for browser or VM, it _might_ work, but it's up to the consumer at the top of the chain (application) to have the final say.

## Changes

- Added a `TransportPlatform` class. This is essentially the same as the `PlatformAdapter` class, but will now be exported publicly so I took the opportunity to rename it.

  A `TransportPlatform` implementation must know how to construct instances of all the different `BaseRequest`s, the `HttpClient`, and the `WebSocket`.

- Added a `BrowserTransportPlatform` class that provides implementations of the transport classes that work in the browser.

- Added a `browserTransportPlatform` constant that is an instance of the aforementioned `BrowserTransportPlatform` class.

- Added a `BrowserTransportPlatformWithSockJS` class that extends `BrowserTransportPlatform` but overrides the `WebSocket` factory to construct an instance that utilizes SockJS instead of native WebSockets.

- Added a `browserTransportPlatformWithSockJS` constant that is an instance of the aforementioned `BrowserTransportPlatformWithSockJS` with default SockJS settings.

- Added a `VMTransportPlatform` class that provides implementations of the transport classes that work on the Dart VM.

- Added a `vmTransportPlatform` constant that is an instance of the aforementioned `VMTransportPlatform` class.

- All `BaseRequest` classes and the `WebSocket` class now accept an optional `TransportPlatform transportPlatform` parameter. This allows injection of the transport platform to be used for individual instances.

- The `HttpClient` class now also accepts an optional `TransportPlatform transportPlatform` parameter. This transport platform will be injected into each request or WebSocket instance that gets created from that client.

- Added a `globalTransportPlatform` variable that acts as a singleton and allows consumers to set the default, global transport platform. If a transport class is explicitly given a `TransportPlatform` instance upon construction, the `globalTransportPlatform` will be used. Setting this is equivalent to using the `configureWTransportForBrowser()` or `configureWTransportForVM()` functions.

- Added an `install()` method to `MockTransports` - this will serve as the new way to configure w_transport for tests (in place of calling `configureWTransportForTest()`).

## Documentation Updates
[...]

PR #224, opened 2016-10-12
CP-2786 Mock updates

evanweible-wf:
_Dependent on #222._

## Description
This PR addresses a couple of issues with the transport mocks:

- There are several mock classes that were exported publicly that are unnecessary and only serve to add confusion.

- Controlling mock websockets involved creating an instance of `MockWSocket` via the transport mocks. This instance would get returned when an actual `WebSocket` instance was constructed, but it also served as the control point for tests, which is confusing.

- The HTTP mock API stores handlers based on the URI, but it was stripping off the query param to make one of our integration tests easier. This really shouldn't have been the case - `expect()` and `when()` are given an expected URI and that URI should match exactly. Now that we have `expectPattern()` and `whenPattern()`, the original use case is much easier so the stripping of query params should no longer be done.

## Changes

- Add a `MockWebSocketServer` class that will be used for controlling the behavior of mock websockets.

- Deprecate any mock class that was publicly exposed but should not be necessary for testing.

- Remove the query param stripping logic from the http mocks.

## Testing

- [ ] CI passes (tests have been added and updated).

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf 
@sebastianmalysa-wf

### DIFF
--- a/lib/mock.dart
+++ b/lib/mock.dart
@@ -24,7 +24,7 @@
 ///     }
 library w_transport.mock;
 
-import 'package:w_transport/src/mocks/transport.dart' show MockTransports;
+import 'package:w_transport/src/mocks/mock_transports.dart' show MockTransports;
 
 export 'package:w_transport/src/http/finalized_request.dart'
     show FinalizedRequest;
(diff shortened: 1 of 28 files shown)


aviary-wf:
## Raven
Number of Findings: 1
* Watched keyword ```withCredentials``` in ```lib/src/http/mock/request_mixin.dart``` line(s) ['81'] added

maxwellpeterson-wf:
+1

maxwellpeterson-wf:
+1 security

evanweible-wf:
@maxwellpeterson-wf added one more commit

maxwellpeterson-wf:
+1

maxwellpeterson-wf:
+1 security

sebastianmalysa-wf:
+1

maxwellpeterson-wf:
QA +1
- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10 with detail of what was tested
- [x] Unit tests created/updated 
- [x] All unit tests pass

Merging into master.

evanweible-wf:

dustinlessard-wf:
+1 (for posterity)

PR #227, opened 2016-10-14
Dont wait for WebSocket to close when the subscription is canceled

evanweible-wf:
## Issue
Currently, `WSocket.cancel()` returns the same `Future` as `WSocket.done`, which means that if you were to wait for cancellation to complete, you would be waiting until the websocket closes.

```dart
final webSocket = await WSocket.connect(...);
final subscription = webSocket.listen((_) { ... });
await subscription.cancel(); // hangs until the websocket closes
```

This is not in line with the behavior that the builtin Dart streams exhibit:

```dart
final sc = new StreamController();
final subscription = sc.stream.listen((_) { .. });
await subscription.cancel(); // completes (doesn't wait for stream to close)
```

## Solution
- Do not wait for the websocket to close when handling the subscription cancellation.
- **Bonus:** also added a test to ensure that the websocket is not closed just because the subscription is canceled. This was added because it's another way in which we should make sure we're in line with `Stream`/`StreamSink` behavior, and because I made a change in 3.0.0-wip that violates this contract, for which I will PR a fix.

## Testing
- [ ] CI passes (tests updated/added)

## Code Review
@trentgrover-wf 
@maxwellpeterson-wf 
@dustinlessard-wf 
@jayudey-wf 
@sebastianmalysa-wf

### DIFF
--- a/test/integration/ws/common.dart
+++ b/test/integration/ws/common.dart
@@ -223,7 +223,7 @@ void runCommonWebSocketIntegrationTests(
     var webSocket = await connect(echoUri);
 
     var subscription = webSocket.listen((_) {});
-    subscription.cancel();
+    await subscription.cancel();
 
     await webSocket.close(4001, 'Closed.');
     expect(webSocket.closeCode, equals(4001));
@@ -244,7 +244,7 @@ void runCommonWebSocketIntegrationTests(
     webSocket.add('one');
     await new Future.delayed(new Duration(milliseconds: 50));
 
-    subscription.cancel();
+    await subscription.cancel();
 
     webSocket.add('two');
     await new Future.delayed(new Duration(milliseconds: 50));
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

test/integration/ws/common.dart
```diff
@@ -254,6 +254,18 @@ void runCommonWebSocketIntegrationTests(
...
+  test('should not close if the only listener is canceled', () async {
+    var webSocket = await connect(echoUri);
+
+    var subscription = webSocket.listen((_) {});
+    await subscription.cancel();
+
+    // Should still be able to add events.
+    webSocket.add('one');
+    expect(webSocket.closeCode, isNull);
+    expect(webSocket.closeReason, isNull);
```
evanweible-wf: Yeah there isn't a good way to make that assertion since I closed the subscription. Asserting that the websocket is still open (via these to expectations here) and implicitly that a `StateError` wasn't thrown should be sufficient.

sebastianmalysa-wf: Awesome, thanks for the reply.

sebastianmalysa-wf:
+1

maxwellpeterson-wf:
+1

maxwellpeterson-wf:
+10
- CI passes

maxwellpeterson-wf:
QA +1
- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10 with detail of what was tested
- [x] Unit tests created/updated 
- [x] All unit tests pass

Merging into master.

PR #265, opened 2017-06-15
Fix failing tests

evanweible-wf:
@toddbeckman-wf


aviary2-wf:
## Raven
Number of Findings: 0

toddbeckman-wf:
+1

PR #279, opened 2017-07-10
WP-4674 Address or ignore deprecated uses

evanweible-wf:
## Issue

There were a bunch of hints from using deprecated members, but most of them are usages of our own deprecated members for the sake of backwards-compatibility.

## Solution

- Switch to non-deprecated alternatives for usages of deprecated members from other packages
- Ignore our own usages (we can address these in the next major)
- Fix a couple of lints that slipped through the cracks

## Testing
- [ ] CI passes

## Code Review
@Workiva/web-platform-pp

### DIFF
--- a/lib/src/http/common/request.dart
+++ b/lib/src/http/common/request.dart
@@ -43,6 +43,7 @@ abstract class CommonRequest extends Object
     autoRetry = new RequestAutoRetry(this);
   }
 
+  // ignore: deprecated_member_use
   CommonRequest.fromClient(Client wTransportClient, client)
       : this._wTransportClient = wTransportClient,
         this.client = client {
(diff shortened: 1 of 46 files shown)


aviary-wf:
## Raven
<details>

</details>

lib/src/http/common/request.dart
```diff
@@ -149,8 +150,9 @@ abstract class CommonRequest extends Object
   /// Whether or not to send the request with credentials.
   bool _withCredentials = false;
 
-  /// [Client] instance from which this request was created. Used in [clone] to
-  /// correctly tie the clone to the same client.
+  /// HttpClient instance from which this request was created. Used in [clone]
```
toddbeckman-wf: The comment claims this is an `HttpClient`, but the code remains a `Client` which isn't in the same class hierarchy.

evanweible-wf: Every instance of client in the codebase right now is typed as `Client`, but is actually an instance of `HttpClient` (this is for backwards compatibility).

You can see this here: https://github.com/Workiva/w_transport/blob/master/lib/src/http/client.dart#L32-L33 – the `Client` factory actually returns an `HttpClient` instance.

toddbeckman-wf:
+1

evanweible-wf:
@jayudey-wf  ready for you

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - passing CI
- [ ] Unit tests created/updated (functionally adding analyzer updates)
- [ ] All unit tests pass

Merging into master.

PR #272, opened 2017-07-04
WP-4632 Global WebSocket monitor

evanweible-wf:
## Description

In order to provide some insight into WebSocket connections at a global level, this PR adds a global "event monitor" for WebSocket connections. This monitor can be used to report analytics.

To obtain a WebSocket event monitor:

```dart
var monitor = WSocket.getGlobalEventMonitor();
```

To listen for connection attempt events:

```dart
var monitor = WSocket.getGlobalEventMonitor();
monitor.didAttemptToConnect.listen((event) {
  // event.wasSuccessful
  // event.url
  // event.sockJsProtocolsWhitelist
  // event.sockJsSelectedProtocol
});
```

## Testing

- [ ] CI pass (tests added)
- [ ] Use this branch to successfully listen for and report analytics on WebSocket connections (instructions coming)

## Code Review

@Workiva/web-platform-pp

PR #281, opened 2017-07-19
WP-4750 Add a global WebSocket monitor (3.x)

evanweible-wf:
## Description

Same as https://github.com/Workiva/w_transport/pull/272, but for the master-line (3.x).

## Testing

- [ ] CI passes
- [ ] Additional testing instructions will be passed to @jayudey-wf 

## Code Review

@Workiva/web-platform-pp

### DIFF
--- a/lib/src/web_socket/browser/sockjs.dart
+++ b/lib/src/web_socket/browser/sockjs.dart
@@ -17,6 +17,7 @@ import 'dart:async';
 import 'package:sockjs_client/sockjs_client.dart' as sockjs;
 
 import 'package:w_transport/src/web_socket/common/web_socket.dart';
+import 'package:w_transport/src/web_socket/global_web_socket_monitor.dart';
 import 'package:w_transport/src/web_socket/web_socket.dart';
 import 'package:w_transport/src/web_socket/web_socket_exception.dart';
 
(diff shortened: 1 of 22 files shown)


aviary-wf:
## Raven
<details>

</details>

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - passing CI
  - verified that event was being emitted as expected when socket was able to connect as well as not connect
- [ ] Unit tests created/updated
- [ ] All unit tests pass

Merging into master.

lib/src/web_socket/browser/sockjs.dart
```diff
@@ -79,12 +81,24 @@ class SockJSWebSocket extends CommonWebSocket implements WebSocket {
     // an error if the socket moves straight to the closed state.
     final connected = new Completer<Null>();
     // ignore: unawaited_futures
-    client.onOpen.first.then((_) => connected.complete());
+    client.onOpen.first.then((_) {
```
dustinlessard-wf: Question : why only the first?

evanweible-wf: This didn't actually change in this PR, it's always been this way, but because there will only ever be one event. The sockJS library just happens to be designed such that all events are dispatched via streams, even if it's a one-time thing

PR #290, opened 2017-09-22
WP-5202 Switch mock-aware requests to real requests before finalizing them

evanweible-wf:
## Problem

When MockTransports are installed with `fallThrough` enabled, we check to see if each request is mock-aware and if it would be handled by an expectation or a mock handler. If it wouldn't be, we switch it to a real request so that it doesn't get stuck (i.e. "fall-through").

Currently, we finalize the request prior to making this check. It turns out that this can have an adverse effect on `MultipartRequests` due to their divergent browser and mock implementations. In the browser, the `MultipartRequest` files map accepts `dart:html.File` and `dart:html.Blob` instances, but those types cannot be imported in any platform other than the browser. As a result, attempting to finalize the mock-aware `MultipartRequest` will fail if either a `File` or `Blob` instance is attached because it does not recognize them.

## Solution

The solution is to switch to a real request (if applicable) _prior_ to finalizing the request, which will avoid the issue altogether if the request isn't destined to be handled by a mock handler or expectation anyway.

## Testing

- [ ] CI passes (applicable test has been updated – you can verify by checking out 76e6488 and running the tests without the fix)

## Code Review

@Workiva/web-platform-pp

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,35 @@
 language: dart
 dart:
   - 1.24.2
-with_content_shell: true
+addons:
+  apt:
+    packages:
+    - ttf-kochi-mincho
+    - ttf-kochi-gothic
+    - ttf-dejavu
+    - ttf-indic-fonts
+    - fonts-tlwg-garuda
 before_install:
+  # Content shell needs this font. Since it has a EULA, we need to manually
+  # install it.
+  #
+  # TODO: remove this and use "sudo: false" when travis-ci/travis-ci#4714 is
+  # fixed.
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Raven
<details>

</details>

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - verified that test was appropriately failing without the corresponding source code update
- [ ] Unit tests created/updated
- [ ] All unit tests pass

Merging into master.

PR #297, opened 2018-01-08
WP-5258 Automatically use sockjs_client_wrapper if possible

evanweible-wf:
## Changes

With the release of [sockjs_client_wrapper](https://github.com/Workiva/sockjs_client_wrapper), we want to use that library if possible because it is fully featured and supports all of the SockJS protocols (particularly xhr-polling). But, we can't just switch because it requires the addition of the `sockjs.js` library to the HTML page.

The proposed solution here is to check for the existence of the `sockjs.js` script by checking for `SockJS` on the window. If it's there, we'll automatically leverage the new package instead of the old Dart port. Otherwise, we'll fallback to the original behavior (i.e. nothing should change).

## Testing

Coming soon..

## Code Review

@Workiva/app-frameworks


aviary-wf:
## Raven
<details>

</details>

PR #299, opened 2018-01-29
Use hasProperty to detect sockjs.js

evanweible-wf:
## Issue

Actually accessing the `SockJS` object on `js.context` throws because it ends up trying to call the object.

## Solution

The proper way to check is to use `js.context.hasProperty` which will simply check to see if the `SockJS` object is available, which is all we're after.

## Code Review
@Workiva/app-frameworks


aviary-wf:
## Raven
<details>

</details>

## Magpie
<details>
<summary>Number of findings: 0</summary>

</details>

example/web_socket/echo/index.html
```diff
@@ -67,6 +74,8 @@ <h1>Echo</h1>
     <pre id="logs"></pre>
 </div>
 
+<!--<script src="packages/sockjs_client_wrapper/sockjs.js"></script>-->
```
jayudey-wf: should these be committed as being commented out?

evanweible-wf: Yeah I don't have a great way to dynamically add/remove the script via the example UI, so I figured this was a good middle ground for now.

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - verified in test case that noticed the bug that this solution was working as expected
  - verified that socket example was still working as exepcted
- [ ] Unit tests created/updated (n/a, dart2js context handling update)
- [ ] All unit tests pass
- [ ] Areas of regression noted

Merging into master.

jayudey-wf:
@Workiva/release-management-pp

PR #301, opened 2018-03-09
Update HttpRequestTypes.md

robertharwell-wf:
Problem: relative path results in broken link. 
Solution:  updated path to explicit
Testing:  Passing Smithy

### DIFF
--- a/docs/guides/HttpRequestTypes.md
+++ b/docs/guides/HttpRequestTypes.md
@@ -1,6 +1,6 @@
 ## HTTP: Request Types
 
-The [basic guide to sending requests](/docs/guides/HttpSendRequestReceiveResponseHandleFailure.md)
+The [basic guide to sending requests](https://github.com/Workiva/w_transport/blob/master/docs/guides/HttpSendRequestReceiveResponseHandleFailure.md)
 introduced the API common to all request classes. Each class, however, is
 designed to transport a specific type of data.
 


robertharwell-wf:
@evanweible-wf - Another relative --> explicit url update.  Please review.

aviary-wf:
## Module Findings
No security relevant content detected. Please review for security relevance and request security review as needed.

docs/guides/HttpRequestTypes.md
```diff
@@ -1,6 +1,6 @@
 ## HTTP: Request Types
 
-The [basic guide to sending requests](/docs/guides/HttpSendRequestReceiveResponseHandleFailure.md)
+The [basic guide to sending requests](https://github.com/Workiva/w_transport/blob/master/docs/guides/HttpSendRequestReceiveResponseHandleFailure.md)
```
evanweible-wf: I'm okay with this because I don't see these guides changing any time soon, but this is another unfortunate limitation of having to use absolute URLs: by pointing to `master`, we will break historical linking. Let's say we release a v4 in the future but someone wants to browse the repo at the latest v3 revision - these links would be broken because they don't allow the user's current context to be taken into account.

evanweible-wf:
+10
- [x] Docs only change

QA +1
@Workiva/release-management-pp

PR #317, opened 2018-07-19
AF-1690 Copy content-type/length on mock-aware fall-through

evanweible-wf:
## Problem

When `MockTransports` are installed, requests are wrapped in a "mock-aware" class when constructed. When `fallThrough: true`, these requests will switch to real requests if there isn't an applicable handler/expectation. This is achieved by constructing a new request instance with the real transport platform on demand and copying all of the request properties over from the mock-aware wrapper to the real request.

We weren't copying the `contentType` and `contentLength` properties. If a consumer had explicitly set one of these properties (rare, but does happen), they would not be copied over in this scenario.

## Solution

Be sure to also copy `contentType` (only if it was explicitly set) and `contentLength` when switching from a mock-aware request to a real request.

## Testing
- [ ] CI passes (test added)

## Code Review
@Workiva/app-frameworks 
@travissanderson-wf

### DIFF
--- a/lib/src/http/common/request.dart
+++ b/lib/src/http/common/request.dart
@@ -124,7 +124,7 @@ abstract class CommonRequest extends Object
   /// Whether or not the content-type was set manually. If `false`, the
   /// content-type will continue to be updated automatically when the [encoding]
   /// is set/changed. If `true`, the content-type will be left alone.
-  bool _contentTypeSetManually = false;
+  bool _wasContentTypeSetManually = false;
 
   /// Completer that should complete when the request has finished (successful
   /// or otherwise).
@@ -191,7 +191,7 @@ abstract class CommonRequest extends Object
   @override
   set contentType(MediaType contentType) {
     verifyUnsent();
-    _contentTypeSetManually = true;
+    _wasContentTypeSetManually = true;
     updateContentType(contentType);
   }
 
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
* Verify [`aviary.yaml`](https://dev.workiva.net/docs/information-security/aviary#configuring-aviary) coverage of security relevant code
---
_Questions or Comments? Reach out on HipChat: InfoSec Forum._

lib/src/http/common/request.dart
```diff
@@ -462,7 +465,7 @@ abstract class CommonRequest extends Object
   /// a real request instance (created from a TransportPlatform instance).
   ///
   /// This is handled by the mock request mixin.
-  Future<BaseResponse> switchToRealRequest({bool streamResponse}) {
```
evanweible-wf: I changed this signature so that I could test `switchToRealRequest` directly.

lib/src/http/mock/request_mixin.dart
```diff
@@ -81,14 +81,22 @@ abstract class MockRequestMixin implements MockBaseRequest, CommonRequest {
...
+    }
+
+    // Encoding cannot be set on MultipartRequests.
     if (this is! MultipartRequest) {
       realRequest.encoding = encoding;
     }
 
-    return streamResponse
-        ? realRequest.streamSend(method)
-        : realRequest.send(method);
```
evanweible-wf: This part is just moved to `CommonRequest` class above.

PR #331, opened 2019-06-26
Prepare for upcoming change to HttpRequest and HttpClientResponse

tvolkert:
An upcoming change to the Dart SDK will change `HttpRequest` and
`HttpClientResponse` from implementing `Stream<List<int>>` to
implementing `Stream<Uint8List>`.

This forwards-compatible change prepares for that SDK breaking
change by casting the Stream to `List<int>` before transforming
it.

https://github.com/dart-lang/sdk/issues/36900

### DIFF
--- a/tool/server/handlers/example/http/cross_origin_file_transfer_handlers.dart
+++ b/tool/server/handlers/example/http/cross_origin_file_transfer_handlers.dart
@@ -125,8 +125,8 @@ class UploadHandler extends Handler {
     final contentType =
         ContentType.parse(request.headers.value('content-type'));
     final boundary = contentType.parameters['boundary'];
-    final stream = request
-        .transform(new MimeMultipartTransformer(boundary))
+    final stream = new MimeMultipartTransformer(boundary)
+        .bind(request)
         .map(HttpMultipartFormData.parse);
 
     await for (HttpMultipartFormData formData in stream) {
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
* Verify [`aviary.yaml`](https://dev.workiva.net/docs/teams/information-security/aviary#configuring-aviary) coverage of security relevant code
---
_Questions or Comments? Reach out on Slack: #support-infosec._

evanweible-wf:
Thanks @tvolkert! We'll need to either drop dart1 support or swap in a manual cast – do you happen to know which SDK release is being targeted for this change? We were already planning on dropping dart1 from our packages relatively soon

tvolkert:
The breaking change will be in Dart 2.5.0, and afaik, `Stream.cast<T>()` was added in Dart 2.0.0, so you'd have to bump the min SDK version to `>=2.0.0`.

tvolkert:
@evanweible-wf I updated this in a way that satisfies the criteria of the fix yet also is Dart 1 compatible.  PTAL?

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: w_transport
-version: 3.2.8
+version: 3.2.9
```
evanweible-wf: Could you back this change out? Our release bot commits the version bump

evanweible-wf:
Awesome, thanks for doing that @tvolkert! Changes look good, if you could revert the pubspec.yaml version bump I can get this merged. Unfortunately we have a git dependency which I think will prevent us from publishing to pub now, and it'll be a breaking change to remove it. I'm prepping that change right now as well and will get it released asap

pubspec.yaml
tvolkert: Done

evanweible-wf:
QA +1
- [x] CI passes and verified that the file transfer example works with these changes

@Workiva/release-management-p

PR #376, opened 2022-07-15
Fix JSON decoding when charset isn't specified on VM to use UTF-8

alanknight-wk:
## Motivation
  
On the VM, JSON responses that do not specify a charset are decoded using latin1/ISO-8859-1. The fallback to latin1 is obsolete behavior, and we should now fall back to the default for the media type. For JSON this is utf-8. On the web, this already behaves correctly, so this should improve consistentency. 

## Changes
  
Hard-code the asJson method on HttpBody to decode as utf-8 from bytes, rather than from string.

## Review
_[See CONTRIBUTING.md][contributing-review-types] for more details on review types (+1 / QA +1 / +10) and code review process._

  

Please review: 

### QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
    - [ ] Steps from PR author: 
        
    - [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
While we perform many automated checks before auto-merging, some manual checks are needed:
- [ ] A Client Platform member has reviewed these changes
- [ ] There are no unaddressed comments _- this check can be automated if reviewers use the "Request Changes" feature_
- [ ] _For release PRs -_ Version metadata in Rosie comment is correct

[contributing-review-types]: https://github.com/Workiva/w_transport/blob/master/CONTRIBUTING.md#review-types
[contributing-manual-testing]: https://github.com/Workiva/w_transport/blob/master/CONTRIBUTING.md#manual-testing-criteria

### DIFF
--- a/lib/src/http/http_body.dart
+++ b/lib/src/http/http_body.dart
@@ -118,7 +118,6 @@ class HttpBody extends BaseHttpBody {
   }
 
   /// Returns this request/response body as a String.
-  ///
   String asString() => _asString(encoding);
 
   String _asString(Encoding charset) {
--- a/lib/src/http/http_body.dart
+++ b/lib/src/http/http_body.dart
@@ -118,12 +118,15 @@ class HttpBody extends BaseHttpBody {
   }
 
   /// Returns this request/response body as a String.
-  String asString() {
+  ///
+  String asString() => _asString(encoding);
+
+  String _asString(Encoding charset) {
     if (_body == null) {
       try {
-        _body = encoding.decode(_bytes);
+        _body = charset.decode(_bytes);
       } on FormatException {
-        throw ResponseFormatException(contentType, encoding, bytes: _bytes);
+        throw ResponseFormatException(contentType, charset, bytes: _bytes);
       }
     }
     return _body;
@@ -137,7 +140,7 @@ class HttpBody extends BaseHttpBody {
   /// than the generic fallback from the response. Throws a [FormatException] if this
   /// request/response body cannot be decoded to text or if the text is not
   /// valid JSON.
-  dynamic asJson() => json.decode((_encoding ?? utf8).decode(asBytes()));
+  dynamic asJson() => json.decode(_asString(_encoding ?? utf8));
 }
 
 /// Representation of an HTTP request body or an HTTP response body where the
--- a/lib/src/http/http_body.dart
+++ b/lib/src/http/http_body.dart
@@ -49,6 +49,7 @@ class HttpBody extends BaseHttpBody {
   String _body;
   Uint8List _bytes;
   Encoding _encoding;
+  Encoding _fallbackEncoding;
 
   /// Construct the body to an HTTP request or an HTTP response from bytes.
   ///
@@ -66,9 +67,9 @@ class HttpBody extends BaseHttpBody {
   /// `charset` param), then [fallbackEncoding] will be used (utf-8 by default).
   HttpBody.fromBytes(this.contentType, List<int> bytes,
       {Encoding encoding, Encoding fallbackEncoding}) {
-    _encoding = encoding ??
-        http_utils.parseEncodingFromContentType(contentType,
-            fallback: fallbackEncoding ?? utf8);
+    _encoding =
+        encoding ?? http_utils.parseEncodingFromContentType(contentType);
+    _fallbackEncoding = fallbackEncoding;
     _bytes = Uint8List.fromList(bytes ?? []);
   }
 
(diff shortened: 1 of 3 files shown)


aviary3-wk:
## Security Insights

## Action Items
* Verify [`aviary.yaml`](https://wiki.atl.workiva.net/pages/viewpage.action?pageId=42762752#SecurityInsightFramework&Tooling-ConfiguringAviary) coverage of security relevant code
---
_Questions or Comments? Reach out on Slack: #support-infosec._

lib/src/http/http_body.dart
```diff
@@ -132,10 +131,11 @@ class HttpBody extends BaseHttpBody {
...
   ///
-  /// This attempts to read this request/response body as a `String` and decode
-  /// it to a JSON object. Throws a [FormatException] if this request/response
-  /// body cannot be decoded to text or if the text is not valid JSON.
-  dynamic asJson() => json.decode(asString());
+  /// This attempts to read this request/response body as a UTF-8 String and
+  /// decode it to a JSON object. Throws a [FormatException] if this
+  /// request/response body cannot be decoded to text or if the text is not
+  /// valid JSON.
+  dynamic asJson() => json.decode(utf8.decode(asBytes()));
```
alanknight-wk, line 138: Possibly this should be use the encoding if it was specified and otherwise utf-8, but my impression is that it's pretty much always utf-8. Like, the HTTP spec says Accept-Charset is deprecated because there's no point.

lib/src/http/response.dart
```diff
@@ -54,8 +53,12 @@ abstract class BaseResponse {
 
   /// Encoding that will be used to decode the response body. This encoding is
   /// selected based on [contentType]'s `charset` parameter. If `charset` is not
-  /// given or the encoding name is unrecognized, [latin1] is used by
-  /// default ([RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html)).
+  /// given or the encoding name is unrecognized, [latin1] is used by default
+  /// ([RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html)).
+  // TODO: That default has been superceded ([RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#appendix-B))
```
todbachman-wf, line 58: Should we create a ticket to address this todo so it doesn't get lost?

alanknight-wk: Created https://jira.atl.workiva.net/browse/FEA-466

lib/src/http/http_body.dart
```diff
@@ -132,10 +132,11 @@ class HttpBody extends BaseHttpBody {
   /// Returns this request/response body as a JSON object - either a `Map` or a
   /// `List`.
   ///
-  /// This attempts to read this request/response body as a `String` and decode
-  /// it to a JSON object. Throws a [FormatException] if this request/response
-  /// body cannot be decoded to text or if the text is not valid JSON.
-  dynamic asJson() => json.decode(asString());
+  /// This attempts to read this request/response body as a UTF-8 String and
+  /// decode it to a JSON object. Throws a [FormatException] if this
```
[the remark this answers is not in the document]
alanknight-wk: Yes. Clarified. It just falls back to utf-8, rather than the response's fallback.

evanweible-wf (approved):

evanweible-wf:
Consumer testing this now..

lib/src/http/http_body.dart
```diff
@@ -132,10 +132,12 @@ class HttpBody extends BaseHttpBody {
...
-  /// This attempts to read this request/response body as a `String` and decode
-  /// it to a JSON object. Throws a [FormatException] if this request/response
-  /// body cannot be decoded to text or if the text is not valid JSON.
-  dynamic asJson() => json.decode(asString());
+  /// This attempts to read this request/response body as a String and decode it
+  /// to a JSON object. If no encoding is specified, it will use UTF-8, rather
+  /// than the generic fallback from the response. Throws a [FormatException] if this
+  /// request/response body cannot be decoded to text or if the text is not
+  /// valid JSON.
+  dynamic asJson() => json.decode((_encoding ?? utf8).decode(asBytes()));
```
brettkail-wk, line 140: This seems to break browsers, which already have the string as `xhr.responseText`, so `asBytes` is converting to latin1 and then this code is attempting to decode with utf8.  Regardless, it seems like string->bytes->string round-trip is undesirable overhead.  Perhaps something like this?
```
String asString() => _asString(encoding);
String _asString(Encoding encoding) { ... } // same as current (using param)
dynamic asJson() => json.decode(_asString(_encoding ?? utf8));
```

evanweible-wf: good catch! @alanknight-wk mind making that update?

alanknight-wk: Oops. Missed this. Done, but I changed the parameter name so as not to shadow the getter.

lib/src/http/http_body.dart
```diff
@@ -118,12 +118,15 @@ class HttpBody extends BaseHttpBody {
   }
 
   /// Returns this request/response body as a String.
-  String asString() {
+  ///
```
evanweible-wf, line 121:
```suggestion
```

evanweible-wf (approved):

evanweible-wf:
QA +1 
- [x] CI passed and consumer tests pass

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #443, opened 2025-07-15
Support setting `Websocket.binaryType`

evanweible-wf:
## Motivation
`WebSocket`s in the browser support a `binaryType` field (see https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType) that we want to allow consumers of our `WebSocket` class to use.

## Changes
Add a `binaryType` field to the `WebSocket` abstraction. It will be a no-op on the Dart VM or when using SockJS, but will work when using native `WebSocket`s.

## Testing/QA Instructions
- [ ] CI passes (tests added)

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -12,4 +12,4 @@ permissions:
 
 jobs:
   publish:
-    uses: Workiva/gha-dart-oss/.github/workflows/publish.yaml@v0.1.7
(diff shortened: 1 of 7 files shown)


travissmith-wf (approved):

travissmith-wf:
QA +1
- tests added to check support for the `arraybuffer` and `blob` binaryType
- CI is passing

rmconsole-wf (approved):
+1 from RM

PR #449, opened 2025-08-26
FEA-8530: Fix late init error

corwinsheahan-wf:
## Motivation
  
We're seeing late initialization errors [here](https://github.com/Workiva/w_transport/blob/master/lib/src/http/browser/utils.dart#L37). 

It seems that pause can get called and execute onPause before onListen executes

## Changes
  
Updated this to be a bit safer and to align more closely with the [Dart-sdk example of how to do a stream transform. ](https://github.com/dart-lang/sdk/blob/2.19.6/sdk/lib/async/stream.dart#L2348-L2381)

This will fix the late init error. 

## Testing/QA Instructions
  
* CI passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+- Removed possibility for a LateInitializationError when transforming a ProgressEvent stream from an HttpRequest into
+a RequestProgress stream.
+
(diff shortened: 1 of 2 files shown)


rmconsole-wf (approved):
+1 from RM

